PR #1, opened 2024-07-15
rename plugins and use workspace to put in common devdeps

Nexucis:


gauge-chart/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@perses-dev/stat-chart-plugin",
+  "name": "@perses-dev/gauge-chart",
```
jgbernalp, line 2:
```suggestion
  "name": "@perses-dev/stat-chart",
```

stat-chart/package.json
```diff
@@ -1,5 +1,5 @@
 {
-  "name": "@perses-dev/gauge-chart-plugin",
+  "name": "@perses-dev/stat-chart",
```
jgbernalp, line 2:
```suggestion
  "name": "@perses-dev/gauge-chart",
```

PR #8, opened 2024-08-06
fix markdown chart name in rsbuild

Nexucis:
It might break things in Perses, but not something that is impossible to fix.

### DIFF
--- a/MarkdownChart/rsbuild.config.ts
+++ b/MarkdownChart/rsbuild.config.ts
@@ -20,19 +20,19 @@ export default defineConfig({
     port: 3005,
   },
   dev: {
-    assetPrefix: '/plugins-dev/Markdown/',
+    assetPrefix: '/plugins-dev/MarkdownChart/',
   },
   output: {
-    assetPrefix: '/plugins/Markdown/',
+    assetPrefix: '/plugins/MarkdownChart/',
   },
   plugins: [pluginReact()],
   tools: {
     htmlPlugin: false,
     rspack: (config, { appendPlugins }) => {
-      config.output!.uniqueName = 'Markdown';
+      config.output!.uniqueName = 'MarkdownChart';
       appendPlugins([
         new ModuleFederationPlugin({
-          name: 'Markdown',
+          name: 'MarkdownChart',
           exposes: {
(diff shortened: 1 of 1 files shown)


MarkdownChart/rsbuild.config.ts
```diff
@@ -20,19 +20,19 @@ export default defineConfig({
...
   plugins: [pluginReact()],
   tools: {
     htmlPlugin: false,
     rspack: (config, { appendPlugins }) => {
-      config.output!.uniqueName = 'Markdown';
+      config.output!.uniqueName = 'MarkdownChart';
       appendPlugins([
         new ModuleFederationPlugin({
-          name: 'Markdown',
+          name: 'MarkdownChart',
```
jgbernalp, line 35: I wanted to match the plugin kind name, so it works correctly with a dashboard definition when being loaded. https://github.com/perses/perses/blob/main/ui/panels-plugin/src/plugins/markdown/markdown-panel-model.ts#L20 We could rename this but we would need to adjust elsewhere so the plugin still works. For example here: https://github.com/perses/perses/blob/feature/plugin-runtime/ui/dashboards/src/components/DynamicPanel/DynamicPanelContent.tsx#L30

Nexucis: yup, that sounds good to me as long you are ok with it as well.

Nexucis: I'm preparing a PR in Perses, reviewing a bit how to install a plugin with the backend involved. Will fix that in the meantime

jgbernalp: I'm ok to be more consistent with the plugin names, but this will mean the current dashboards using `Markdow` need to be migrated to use `MarkdownChart`, right?

jgbernalp (approved):

MarkdownChart/rsbuild.config.ts
Nexucis: yes migration will be needed

Nexucis: that's definitively a breaking change

PR #11, opened 2024-11-05
Add CUE files for panel plugins

AntoineThebaud:
- Add CUE files for all panel plugins. Copied from [perses/perses](https://github.com/perses/perses) at revision `cdd8360987e5051ad056ffa506386fa0dbe4d4fd`
- Make each plugin a standalone CUE module
   - Vendor/copy the common package dependency for each plugin
- Copy also the plugins-specific doc

This is a first step of the plan I described here https://github.com/perses/perses/discussions/2085#discussioncomment-11078190

### DIFF
--- a/BarChart/backend/bar.cue
+++ b/BarChart/backend/bar.cue
@@ -1,0 +1,26 @@
+// Copyright 2023 The Perses Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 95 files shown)


Nexucis:
I am wondering if we really want to version the vendor deps from perses/perses ? or instead having a sort of deps files and using the CLI it will install what is necessary ?

AntoineThebaud:
that's what I was thinking to do with my ongoing work on new percli commands, but let's say following the "legacy" modules & dep management of CUE this is how you should manage the deps "by the book":

```
cue.mod
|-- ...
|-- pkg         // copies of external packages
...
```
(from https://cuelang.org/docs/concept/modules-packages-instances/)

I like the idea to do a first PR that follows that, and then improve the situation incrementally (next step is using custom logic from Perses, final step would be to rely on the new dep management of CUE)

PR #12, opened 2024-11-18
CI: add new step for CUE schemas validation

AntoineThebaud:


scripts/schemas/validate.go
```diff
@@ -0,0 +1,58 @@
...
+	// check for errors on the instances
+	if buildInstance.Err != nil {
+		return buildInstance.Err
+	}
+
+	return nil
+}
+
+func main() {
+	workspaces, err := npm.GetWorkspaces()
```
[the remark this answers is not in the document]
jgbernalp: should we create a script/function that can be reused per plugin. Would this be helpful to be included in the plugin template so eternal implementations can also benefit from this?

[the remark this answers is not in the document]
Nexucis: I would keep it this way. A plugin without the frontend part cannot work. So getting the list of the npm packages to list the plugin here is ok from my opinion.

By looking at the (already big) list of the plugins, I am wondering if we shouldn't create a sub folder `panel`. Just to avoid having everything at the root and not able to know by a quick look where are the panels / datasources / queries ...etc.

scripts/validate-schemas/validate-schemas.go
```diff
@@ -0,0 +1,58 @@
...
+import (
+	"fmt"
+	"path/filepath"
+
+	"cuelang.org/go/cue/load"
+	"github.com/perses/perses-plugins/scripts/npm"
+	"github.com/sirupsen/logrus"
+)
+
+const schemasPath = "backend"
```
Nexucis, line 25: why not calling the folder `schema` instead ? backend for me doesn't define the place where the schema should be defined

scripts/schemas/validate.go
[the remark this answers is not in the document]
Nexucis: > should we create a script/function that can be reused per plugin. Would this be helpful to be included in the plugin template so eternal implementations can also benefit from this?

Yeah that would make sense definitely. I would add it in the CLI for sure

scripts/validate-schemas/validate-schemas.go
AntoineThebaud: well I called it that way because it's technically the backend part of the plugin. I dont want to change the structure of the frontend side while gabriel is working on it but I was thinking that in the end a plugin could look like
```
MyPlugin/
- backend/
  - cue.mod/
  - mypluginschema.cue
  - migrate.cue
- frontend
  - package.json
  - src/
  - ...
```
Or maybe that's unnecessary?

AntoineThebaud: well the thing is, the day we'll be ready to implement the plugin installation part, schemas & frontend code need to be installed in different places probably? that's why I was thinking to have wrapper folders here to separate them, to hopefully make this easier.

Nexucis: Actually, in the current (new) implementation of the plugin management in Perses, schema & frontend can be in the same place and it will make sense they are in the same places actually

AntoineThebaud: Tackled offline, this is actually not how the plugin files were planned to be organized, let's change that in another PR since this is not related to that one

PR #13, opened 2024-11-19
Remove vendored CUE dependencies

AntoineThebaud:
Replace the vendored CUE dependencies by a call to `percli plugin update`
- for local dev env: add doc to explain how to setup things
- for CI: new job to get the dependencies, that gets triggered before running the validation & building the plugin archives (⚠️ cue files not embedded in the plugin archives yet)


Nexucis (approved):
nice !

PR #16, opened 2024-12-16
Add prometheus datasource plugin

jgbernalp:
This PR has several changes:

- Refactors the `perses` field in the package.json to hold information regarding the Plugin Module, this allows that a single plugin package can hold several types of plugins. Useful for datasources plugins which can contain datasource, variables and queries.
- Adds the Prometheus plugin
- Fixes the warning issue regarding non compatible version found by providing a required shared version
- Adjusts the build archive script to be friendly with unix systems


Nexucis:
That's awesome @jgbernalp ! Thank you for doing that ! 🙏 

To fix some of the CI issues, you should delete the file `prometheus/package-lock.json` and then run npm install at the root of the repo. It will fix some `npm` issue as we are in an npm workspace and as such, you should have only one `package-lock.json`

Prometheus/src/components/PromQLEditor.tsx
```diff
@@ -0,0 +1,160 @@
...
+  }
+  return errorMessage;
+}
+
+export type PromQLEditorProps = {
+  completeConfig: CompleteConfiguration;
+  datasource: PrometheusDatasourceSelector;
+} & Omit<ReactCodeMirrorProps, 'theme' | 'extensions'>;
+
+export function PromQLEditor({ completeConfig, datasource, ...rest }: PromQLEditorProps) {
```
Nexucis, line 45: it looks like you don't have the last version of the plugin. Looking at the branch `main` in  `perses/perses`, every function has the returned type now. So I am wondering if there are other things/features missing.

Nexucis: At some point we need to freeze every changes regarding the plugin in the `perses/perses` repo to avoid the desynch

jgbernalp: yeah I realized that when merging Perses main. I'll update the plugin. 

Yeah I think is unavoidable to have changes on both sides until we stabilize the API. We can maybe add a renovate rule to keep the plugin branch updated.

Nexucis: I believe we are quite closed to say "let's freeze the update of the plugin on perses/perses side."

I supposed once this PR is merged and can loaded in Perses, then we can start to migrate the plugin right ?

Nexucis:
The validation of the cuelang schema is failing. I think it's ok to ignore this issue and to merge this PR to move forward on this topic.

I need to think a bit if we should make a quick and dirty fix in the CI or if we should start to use the `plugin` command in the CLI and to update them accordingly

Prometheus/package.json
```diff
@@ -0,0 +1,72 @@
...
+        },
+        "kind": "PrometheusDatasource",
+        "pluginType": "Datasource"
+      },
+      {
+        "display": {
+          "name": "Prometheus Time Series Query"
+        },
+        "kind": "PrometheusTimeSeriesQuery",
+        "pluginType": "TimeSeriesQuery"
```
Nexucis, line 47: `TimeSeriesQuery` or simply `Query` ? On the backend side, we don't make distinction about the different type of query for the moment.

Do you need this distinction in the frontend @jgbernalp ?

jgbernalp: Yes, I think we should discuss about making that distinction on the backend as well or add these base interfaces as a core plugin. This will serve as common interfaces between datasource data and charts. For example in the frontend we currently have TimeSeriesQuery and [TracesQuery](https://github.com/perses/perses/blob/main/ui/plugin-system/src/model/plugins.ts#L63).

Prometheus/src/components/PromQLEditor.tsx
jgbernalp: Yes, this is the PR that loads this plugins in Perses: https://github.com/perses/perses/pull/2514, I'm currently migrating the rest of the plugins.

Nexucis:
@jgbernalp I have fixed the rest of the CI issue. I think we are good to merge.

Prometheus/package.json
Nexucis: yeah ok, let's discuss about it in a call sometimes

AntoineThebaud: Linking back the discussion around this topic https://github.com/perses/perses/discussions/1922, but y we probably need to think about all the potential implications that "hacking" the plugin kind (to put instead what we could called a plugin "specialized" kind) would have, before going that way.

jgbernalp: I think in this particular case the problem is the frontend definition of the PluginModule. In reality kind will be TimeSeriesQuery and pluginType will be the actual name of the plugin. Following the kind,  metadata, spec format.

```json
{
  "kind": "PluginModule",
  "metadata": {
    "name": "BarChart"
  },
  "spec": {
    "plugins": [
      {
        "kind": "Panel",
        "metadata": {
          "name": "BarChart"
        },
        "spec": {
          "display": {
            "name": "Bar Chart"
          }
        }
      }
    ]
  }
}
```

PR #19, opened 2024-12-23
put back list of files in the final build archive and add cue.mod folder

Nexucis:
Putting back in the archive a list of files and folders require to be a valid plugin.

I am also adding the the folder `cue.mod` and `schemas` that will be required later in the backend to allow the plugin in the database.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -57,7 +57,7 @@ jobs:
           enable_go: true
       - uses: ./.github/perses-ci/actions/install_percli
         with:
-          cli_version: "main-2024-11-13-e99df92a-distroless" # To replace by latest when `percli plugin update` command will be released
+          cli_version: "v0.50.0-rc.0"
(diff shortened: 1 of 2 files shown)


jgbernalp:
We can control this also from the rspack side when packaging the plugin. I did it [here](https://github.com/perses/plugins/blob/main/TimeSeriesChart/rsbuild.config.ts#L27) so is consistent when you are developing plugins from a node environment without the go package script. WDYT?

Nexucis:
ah that's interesting, I didn't see this part. 

For sure in an environment where go doesn't exist it can be interesting. In my head (currently), this script will be replaced by the usage of the perses CLI with the command [`percli plugin build`](https://github.com/perses/perses/blob/main/internal/cli/cmd/plugin/build/build.go) that will take care about the files required in the archive.

Also like that we don't need to duplicate the list of files to add to archive in every rspack config.

For the moment I need to adjust the CLI, there are some issues with this command. And it will be provide to external people a single way to build their plugin.

PR #25, opened 2025-01-15
add remaining plugins, refactor exports and add test and lint commands

jgbernalp:
this huge PR 😅. Adds the following:

- Migrates the remaining plugins:

  - BarChart-0.4.0
  - GaugeChart-0.4.0
  - MarkdownChart-0.4.0
  - PieChart-0.4.0
  - Prometheus-0.4.0
  - ScatterChart-0.4.0
  - StatChart-0.4.0
  - StaticListVariable-0.1.0
  - Table-0.4.0
  - TimeSeriesChart-0.4.0
  - TimeSeriesTable-0.4.0
  - TraceTable-0.4.0
  - TracingGanttChart-0.4.0

- Changes the way the plugin is exported to match better the editing panels and options
- bumps the version to 0.4.0 as the export changes are breaking changes
- Adds test and lint commands
- Fixes the issue of the panel and datasources editors

Review will be probably better on a small meeting.

This needs https://github.com/perses/perses/pull/2557


scripts/build-archive/build-archive.go
```diff
@@ -44,21 +46,38 @@ func createArchive(pluginName string) error {
...
+	if runtime.GOOS == "darwin" {
+		args = append([]string{"--disable-copyfile"}, args...)
+	}
+
+	if execErr := exec.Command("tar", args...).Run(); execErr != nil {
 		return execErr
 	}
 
+	if createGroupArchive {
+		if execErr := exec.Command("cp", path.Join(pluginName, archiveName), path.Join("./plugins-archive", archiveName)).Run(); execErr != nil {
```
Nexucis, line 61: Maybe I forgot something, but here I don't get where the binary `plugins-archive` is coming from. I can admit I don't even get why we would need that

Nexucis:
increadible work @jgbernalp !!! 🤯  I just have a small comment and I was wondering when you are saying:

> This needs https://github.com/perses/perses/pull/2557

Does that mean we need to merge https://github.com/perses/perses/pull/2557 before this one ? Which is odd for me as the CI is green here and red in the other one.

Nexucis:
> Review will be probably better on a small meeting.

I think we can move forward and merge this PR and make this meeting afterward.

scripts/build-archive/build-archive.go
jgbernalp: It was just for convenience while developing locally to replace it in the Perses side. But we can remove it once we can load a development version of the plugin files from the `/plugins/` endpoint in Perses.

jgbernalp:
> increadible work @jgbernalp !!! 🤯 I just have a small comment and I was wondering when you are saying:
> 
> > This needs [perses/perses#2557](https://github.com/perses/perses/pull/2557)
> 
> Does that mean we need to merge [perses/perses#2557](https://github.com/perses/perses/pull/2557) before this one ? Which is odd for me as the CI is green here and red in the other one.

Yes. Is green here because the test of the plugins in isolation are ok. But we don't have integration tests yet.

scripts/build-archive/build-archive.go
Nexucis: ah yeah makes sense. We will need to provide something quickly, it will be super annoying otherwise to develop a plugin

PR #37, opened 2025-02-03
add explorer plugins

jgbernalp:
This PR:

- Adds the explorer views for Prometheus and Tempo plugins.

Needs changes from Perses packages, in particular exports from `@perses-dev/explore` https://github.com/perses/perses/pull/2598


Prometheus/package.json
```diff
@@ -11,26 +11,32 @@
...
+    "color-hash": "^2.0.2",
+    "qs": "^6.13.0"
   },
   "peerDependencies": {
     "@emotion/react": "^11.7.1",
     "@emotion/styled": "^11.6.0",
     "@hookform/resolvers": "^3.2.0",
     "@perses-dev/components": "^0.50.1",
     "@perses-dev/core": "^0.50.1",
+    "@perses-dev/dashboards": "^0.50.1",
```
Nexucis, line 25: I guess ideally explorer shouldn't be dependent on the package `@perses-dev/dashboards` that sounds heavy.

WDYT @jgbernalp @celian-garcia @Gladorme ?

jgbernalp: Yes is heavy. The main component used here is the Panel, which we might move to @perses-dev/components. I Just wanted to do minimal changes so this can work.

jgbernalp: For the moment this will be provided by Perses UI as a shared dependency

jgbernalp:
While testing this, I found an issue. Here the fix https://github.com/perses/perses/pull/2607

Nexucis (approved):
🥳

Prometheus/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@perses-dev/prometheus",
   "private": true,
-  "version": "0.4.1",
+  "version": "0.4.2",
```
Nexucis, line 4:
```suggestion
"version": "0.5.0",
```

Because you are adding new functionalities, it's not a simple patch

Tempo/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "name": "@perses-dev/tempo",
   "private": true,
-  "version": "0.1.0",
+  "version": "0.1.1",
```
Nexucis, line 4:
```suggestion
"version": "0.2.0",
```
same reason than for Prometheus, it's not a patch

PR #39, opened 2025-02-05
add go implentation for barchart

Nexucis:
Here a proposition to move the golang implementation of the BarChart.

To be installed like that:

```bash
go get github.com/perses/plugins/barchart
```

I am just wondering if it's fine to have the go files at the root of the package in the middle of the frontend files.

Any thoughts on that @jgbernalp @AntoineThebaud ?


Nexucis:
once we have an agreement on how it should be handled, I will create another PR with the rest of the plugins

jgbernalp:
should we then move the frontend to a `ui` folder? I wonder if we would have plugins only for backend

AntoineThebaud:
I know it's not needed to exclude these files for the plugin archive build, but wouldnt it be clearer to gather all them under a `go-sdk` folder? The day we support more SDKs it's probably how we'd like to separate things I guess, but I have a small preference to already do that now

Nexucis:
I have moved the go files to the folder `sdk/go` I think it makes more sense that the UI files stay at the root of the plugin.

jgbernalp (approved):

PR #42, opened 2025-02-05
[BUGFIX] GaugeChartPanel: fix empty state

andreasgerstmayr:
Ref https://github.com/perses/perses/pull/2548

The `GaugeChart` only checks for `undefined`, not `null`, when deciding whether to show the no data display:
https://github.com/perses/perses/blob/c4a7621b138b903599743c7162526a5cbc7cb73e/ui/components/src/GaugeChart/GaugeChart.tsx#L52


Nexucis (approved):
ooups. Thank you @andreasgerstmayr for fixing it twice !

PR #44, opened 2025-02-05
add local plugin development instructions to README

andreasgerstmayr:
Add local plugin development instructions to README

ref. https://github.com/perses/perses/pull/2593

### DIFF
--- a/README.md
+++ b/README.md
@@ -7,3 +7,20 @@ This repository contains the core plugins for [Perses](https://github.com/perses
 ### Get plugin schemas dependencies
 
 Most if not all the plugins schemas rely on the [`common` CUE package of the perses/perses repo](https://github.com/perses/perses/tree/main/cue/schemas/common). To be able to evaluate the CUE model of a given plugin locally, you thus need to make this package available as en external CUE dependency for it. In the future this is something where you'd just have to run `cue get` (in the same way as `go get` for Golang libs), but for now you have to rely on an utility we provide through percli: `percli plugin update`.
+
+## Development
+
+The instructions are for the `Table` plugin, replace the name accordingly.
+
+1. Start development server of the plugin: `cd Table; npm run dev`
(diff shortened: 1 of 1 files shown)


andreasgerstmayr:
It would be cool to start a single server for all plugins, but when I try to register `ModuleFederationPlugin` twice in `rsbuild.config.ts`, I get an error:
```
error   Detect duplicate register RspackModuleFederationPlugin,please ensure RspackModuleFederationPlugin is singleton!
```

I could add multiple entries to `exposes`, but `name` is a single string, not sure how that would work.

README.md
```diff
@@ -7,3 +7,20 @@ This repository contains the core plugins for [Perses](https://github.com/perses
...
+   ```
+   plugins:
+     dev_environment:
+       plugins:
+         - name: Table
+           url: http://localhost:3005
+           absolute_path: /absolute/path/to/plugin/repository/Table
+   ```
+1. Start Perses backend (in `perses` repository): `./scripts/api_backend_dev.sh`
+1. Start Perses frontend (in `perses` repository): `cd ui; npm run start`
```
Nexucis, line 26: I am not sure you need to do this step. The frontend won't reload automatically once the plugin files are changing, you will need to reload manually the Perses UI.

The UI is available at http://localhost:8080 after starting the backend

jgbernalp:
I think the way to have a dev server with all the plugins will be to start them in different ports and build a proxy. Basically the same as [turbopack](https://turbo.build/repo/docs/crafting-your-repository/creating-an-internal-package) does, maybe we can create a dev internal package that references all the plugin as internal dependencies so a single server can be started.

README.md
jgbernalp: Yes but we could add https://github.com/module-federation/fmr so Perses UI hot reloads the plugin when it changes

Nexucis: oh that would be cool indeed !

Nexucis (approved):

PR #46, opened 2025-02-07
[IGNORE] TimeSeriesChart: mystery fix for perses/perses migration unit tests 

AntoineThebaud:
The previous code with the switch-like statement is strictly equivalent to new one, and it is working fine with Perses "live", no issue to migrate dashboards from the UI etc. but for some reason it is causing this error in the migrate unit tests with the bump to cue v0.12 on perses/perses side: `cue: marshal error: spec.visual.lineWidth: cannot convert incomplete value "_" to JSON.`

Ref https://github.com/perses/perses/pull/2599

### DIFF
--- a/TimeSeriesChart/package.json
+++ b/TimeSeriesChart/package.json
@@ -1,7 +1,7 @@
 {
   "name": "@perses-dev/timeseries-chart",
   "private": true,
-  "version": "0.4.1",
+  "version": "0.4.2",
   "scripts": {
     "dev": "rsbuild dev",
     "build": "rsbuild build",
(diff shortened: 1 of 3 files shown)


TimeSeriesChart/cue.mod/module.cue
```diff
@@ -1,4 +1,4 @@
 module: "github.com/perses/plugins/timeserieschart@v0"
 language: {
-	version: "v0.11.0"
+	version: "v0.12.0"
```
Nexucis, line 3: probably we should bump the version for all plugin

AntoineThebaud: cue version upgrade for all plugins to be tackled in a separate PR as agreed

Nexucis (approved):

Nexucis:
can you please increment the version number in the package.json of the plugin and then run `npm install` at the root of the repo ?

Thanks

PR #52, opened 2025-02-12
downgrade promql code mirror to avoid warning, add keys to tooltip buttons

jgbernalp:
This PR:

- downgrades the `@prometheus-io/codemirror-promql` to avoid the [auto purge warning](https://github.com/prometheus/prometheus/blob/906f6a33b60cec2596018ac8cc97ac41b16b06b7/web/ui/module/codemirror-promql/src/client/prometheus.ts#L297)
- Adds keys to the buttons on tooltips to make react happy

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -2852,6 +2852,35 @@
         "url": "https://opencollective.com/popperjs"
       }
     },
+    "node_modules/@prometheus-io/codemirror-promql": {
+      "version": "0.45.6",
+      "resolved": "https://registry.npmjs.org/@prometheus-io/codemirror-promql/-/codemirror-promql-0.45.6.tgz",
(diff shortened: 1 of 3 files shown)


Nexucis (approved):

Nexucis:
While I agree to downgrade, I don't get why there is an impact with codemirror as it is an internal dependencies to the prometheus module. It shouldn't be an issue with Perses right ?

Besides how can we upgrade this dependencies then ? (because I know it contains many fixes about the promQL autocompletion)

jgbernalp:
Is not an issue per se, is reported on the console and the e2e consider that as an error. It should be a warning. We need to check if warnings are being reported as errors from the plugin side to be able to upgrade, and maybe contribute upstream to allow to provide this option as configuration.

jgbernalp:
This can be updated to the latest version when https://github.com/prometheus/prometheus/pull/16024 is included in a release

PR #55, opened 2025-02-13
CUE: rely on native dep management

AntoineThebaud:
(targetting feature branch)

Get rid of the usage of `percli plugin update` command and rely instead on CUE's native dependency management.

`perses/perses` schemas published to the Central Registry: https://registry.cue.works (manual POC for now)

> [!WARNING]  
> With this PR the deps of the plugin are no longer packaged with it. Letting things like this would imply that the Perses server should have access to an OCI registry - thus either access to internet OR users would have to setup a registry in their private network themselves. As we consider this constraint a no-go in our case, I was looking for solutions to that problem... After asking on CUE's slack it seems our long term solution would be this https://github.com/cue-lang/cue/issues/3328.
>
> For the time being, to work around this problem we'll have to tune the `percli plugin build` command to do this before building the archive:
> 1. set the `CUE_CACHE_DIR` env var to e.g `./temp`
> 2. run `cue eval`
> 3. run `mv $CUE_CACHE_DIR/* ./cue.mod/pkg/`
> 4. edit the `cue.mod/module.cue` of the plugin to remove the `deps` part -> as it is a .cue file we know how to parse + modify + serialize back with the modification, so "no so dirty" in the end..


barchart/cue.mod/module.cue
```diff
@@ -2,3 +2,12 @@ module: "github.com/perses/plugins/barchart@v0"
 language: {
 	version: "v0.11.0"
 }
+source: {
+	kind: "git"
+}
+deps: {
+	"github.com/perses/perses/cue@v0": {
+		v:       "v0.0.1-test"
```
AntoineThebaud, line 10: version to be changed when it'll be time to merge on main.
We'll have to align versioning of the module with the versioning of the binary. The same should apply to the versioning of the plugin schemas too.

PR #59, opened 2025-02-22
[DOC] Update README.md for plugin local development

azizthanawala:
Hello,

I felt it would be helpful to add this little caveat I found while setting up the plugins repository for local development.
Not sure if the correct catalog-entry would be DOC or IGNORE in this case so please suggest if otherwise.


AntoineThebaud (approved):
Thanks!

PR #64, opened 2025-02-28
Use turbo to run tasks concurrently

andreasgerstmayr:
* use turbo to run all tasks concurrently
* replace devserver with `turbo run dev`, i.e. starting all dev servers in parallel, because running multiple rsbuild dev servers in the same process unfortunately doesn't work reliably (see https://github.com/web-infra-dev/rspack/issues/9538 for details).
* give unique port numbers to all dev servers


andreasgerstmayr:
@AntoineThebaud the CI is failing on the cue login step, could you take a look?

Nexucis:
yeah that is the same issue than in `perses/perses`. See also my comment here: https://github.com/perses/plugins/pull/80#issuecomment-2715356587

We can merge this one.
The only thing I can suggest for the moment is to avoid using a fork and instead push in a branch directly in the upstream. Let's Friday what will be the result of the meeting

PR #67, opened 2025-03-03
[FEATURE] publish npm packages

jgbernalp:
This PR includes these changes for all the plugins:

- script that publishes the plugin as an npm package.
- Updates the rspack federation dependency version and configuration
- Compiles ESM, CJS and federation assets into the `dist` folder
- Bumps a new version, in the case of prometheus and tempo uses the `0.51.0-beta.2` version as these plugins already exist in npm


prometheus/package.json
```diff
@@ -1,7 +1,6 @@
 {
-  "name": "@perses-dev/prometheus",
-  "private": true,
-  "version": "0.5.1",
+  "name": "@perses-dev/prometheus-plugin",
```
Nexucis, line 2: I am wondering if we shouldn't keep `@perses-dev/prometheus` to show the changes and mark the previous package deprecated.
WDYT ?

jgbernalp: Yeah I thought about it. But then the packages won't have the plugin name on it. At the end they have the same content but in a breaking change fashion 😜 . Should use the `@perses-dev/{plugin_name}-plugin` format also for other plugins?

Nexucis: well if we want to go this way, yeah I guess we need to have this naming convention to make the difference with the core package.

but that means next time for prometheus and tempo we will have to tag them with the version v0.51.0 insteand of v0.6.0 or equivalent

.github/workflows/ci.yml
```diff
@@ -139,4 +139,7 @@ jobs:
         uses: actions/download-artifact@v4
         with:
           name: archives
-      - run: go run ./scripts/upload-archive/upload-archive.go -tag=${{ github.event.release.tag_name }}
+      - name: Upload archives to GitHub releases
+        run: go run ./scripts/upload-archive/upload-archive.go -tag=${{ github.event.release.tag_name }}
+      - name: Publish npm package
+        run: go run ./scripts/npm/npm-publish.go -tag=${{ github.event.release.tag_name }}
```
AntoineThebaud, line 145:
```suggestion
        run: go run ./scripts/npm-publish/npm-publish.go -tag=${{ github.event.release.tag_name }}
```

AntoineThebaud, line 145: I guess you have to change `enable_npm` to true above?

scripts/npm-publish/npm-publish.go
```diff
@@ -0,0 +1,48 @@
...
+	"os/exec"
+	"path/filepath"
+	"regexp"
+
+	"github.com/perses/plugins/scripts/npm"
+	"github.com/sirupsen/logrus"
+)
+
+func main() {
+	tagNamePattern := regexp.MustCompile(`(?m)(.+)/v(\d.\d.\d)`)
```
Nexucis, line 28: I guess every script running for the release will need this regexp. It will be worth to have a common package where there is this pattern and the function returning the plugin folder name and the version.

scripts/npm-publish/npm-publish.go
```diff
@@ -0,0 +1,48 @@
...
+	pluginFolderName := tagSplitted[1]
+	version := tagSplitted[2]
+	// The manifest is hopefully uploaded by a previous task in the CI
+	// It should be available in the plugin folder
+	manifest, err := npm.ReadManifest(pluginFolderName)
+	if err != nil {
+		logrus.WithError(err).Fatalf("unable to read manifest file for plugin %s", pluginFolderName)
+	}
+	pluginName := manifest.Name
+	if execErr := exec.Command("npm", "publish", "--access", "public", filepath.Join(pluginFolderName, fmt.Sprintf("%s-%s.tar.gz", pluginName, version))).Run(); execErr != nil {
```
Nexucis, line 45: this looks weird to me to publish on npm the `tar.gz` archive built. I am think about when you are importing the package. Usually it's done thanks to what is written in package.json giving the entrypoint and it is built with `tsc`.

How will it work with the static part ?

Nexucis: oh ok, I didn't know that. Thanks for the explanation !

scripts/npm-publish/npm-publish.go
AntoineThebaud: [I confirm](https://github.com/perses/plugins/pull/70) 😄

AntoineThebaud: Hmm actually in the case of CUE module the regexp should be different ("v" must be kept as well as pre-release suffix if provided), so not sure if we can/should factorize stuff here

Or, we make the regexp more complex with more matching groups & subsequent scripts take only what they're interested in

Nexucis: the regexp here doesn't care about the pre-released suffix because I forgot about it when I write it. But that should be considered for sure.

Regarding the `v`, I think if you have the correct version, then adding `v` to the version won't be that hard.

I have opened the PR https://github.com/perses/plugins/pull/71 to put in common the logic to get and parse the flag

Nexucis:
Once the PR is updated with the changes provided in https://github.com/perses/plugins/pull/71, we can merge this one

sjcobb:
> Once the PR is updated with the changes provided in #71, we can merge this one

👋 Just following along, I think this was completed now right, any other open items? Once plugins are published, I'd be happy to help update the embedding-panels docs / create a new embed example

jgbernalp:
@AntoineThebaud @Gladorme This is good to go now. PTAL

Gladorme (approved):
lgtm 😄

AntoineThebaud (approved):
LGTM, just a minor thing

PR #72, opened 2025-03-10
move dac doc about prometheus to the plugins

Nexucis:


AntoineThebaud (approved):
👍 I wonder if we wouldnt rather have the plugin-specific doc moved to the corresponding plugin folder, but let's go like that for now anyway.
Btw there are other remaining dac docs to move then

Nexucis:
> I wonder if we wouldnt rather have the plugin-specific doc moved to the corresponding plugin folder

I have no opinion, I am fine either way

PR #75, opened 2025-03-10
[IGNORE] CI: Build plugins using percli

AntoineThebaud:


Nexucis:
also when the command failed to vendor the dependencies, it seems the command does not remove the folder `pkg.bak`

Nexucis:
I have prepared a fix to cover the issues mentioned above: https://github.com/perses/perses/pull/2748. I will add more test in this PR hoping it will be the last PR to finally be able to build the plugins

Nexucis (approved):
with the latest changes from perses, it's finally working.

Probably we need to adapt the load of the plugin in Perses when they are in development so Perses can download the dependencies as well

PR #77, opened 2025-03-11
[DOC] add DaC docs for staticlistvariable

AntoineThebaud:
moved from perses/perses.

The Go doc still has to be enriched, to be tackled later


Nexucis (approved):
❤️

PR #80, opened 2025-03-11
Migrate all panels to receive data via prop, update Perses dependency

andreasgerstmayr:
* migrate all panels to receive data via prop
* update Perses dependency to 0.51.0-beta.0
* update @typescript-eslint/eslint-plugin to match typescript version


Nexucis:
there is again the issue with the cuelang token required ...

With Antoine we will have a meeting this friday with Cuelang to talk about this. I hope we will be able to convince them to remove this requirement. It's blocking all pipeline we have

PR #84, opened 2025-03-12
[ENHANCEMENT] Table: support custom table cell height

andreasgerstmayr:
We already support custom table cell width, this PR adds changing the cell height. This is required when we show panels in table cells.

I also updated the switch component to reflect that the default value is `auto` (i.e. if value is `undefined`, the switch should be enabled).

Requires: https://github.com/perses/perses/pull/2762

## Screenshots

Default:
Bildschirmfoto vom 2025-03-12 17-01-51

Custom height:
Bildschirmfoto vom 2025-03-12 17-02-09


Gladorme (approved):
LGTM 😄

PR #88, opened 2025-03-17
[IGNORE] Use separate CI pipeline for PRs due to missing CUE token on PRs

andreasgerstmayr:
CUE Central Registry login does not work on forks because GitHub Secret is not available.

cc @AntoineThebaud @Nexucis

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,7 +3,6 @@ on:
   push:
     branches:
       - main
-  pull_request:
   merge_group:
   release:
     types:
--- a/.github/workflows/pr.yml
+++ b/.github/workflows/pr.yml
@@ -1,0 +1,56 @@
+name: ci
+on:
+  pull_request:
+
+jobs:
+  build:
+    name: 'build'
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@v4
+      - uses: perses/github-actions@v0.9.0
+      - uses: ./.github/perses-ci/actions/setup_environment
+        with:
+          enable_npm: true
+      - run: npm ci
+      - run: npm run build
+
+  lint:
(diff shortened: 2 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -12,6 +12,8 @@ on:
 jobs:
   build:
     name: 'build'
+    # CUE Central Registry login does not work on forks because GitHub Secret is not available
+    if: ${{ ! github.event.pull_request.head.repo.fork }}
```
Nexucis, line 16: yeah but the issue by doing that, is that we don't know anymore if the plugin can be built

Nexucis: that's not true, we changed the build of the plugin and now the CLI is running the frontend build

Nexucis: but yes the tests for the frontend are still running that is true

andreasgerstmayr: hm, yes, iirc `jest` compiles just-in-time, i.e. only files which are required by tests. then we only have type-check left :\

so, what should we do? Is the CUE registry dropping the requirement of needing a token anytime soon?

Nexucis: we should have an answer this week, but for the moment I have no good solution :(

andreasgerstmayr: or maybe we could cache the cue dependencies? we don't need to download them every time from the central registry. of course this breaks when a fork is updating the dependencies, but I think that won't happen as much

Nexucis: ah yeah that's an idea that will mitigate the dependency from the Cuelang team decision.

Nexucis:
yeah let's go like that. It will release a bit the pressure on finding a fine solution for cue.

Nexucis (approved):

PR #90, opened 2025-03-18
cache cue deps

Nexucis:
trying the idea from @andreasgerstmayr in PR #88.

Let's see if it's working


Nexucis:
I am not sure the cache is working. For the moment it doesn't save anything. I am wondering if it's because it's in a PR, and the cache is only saved on the main branch.

Nexucis:
in the `build` job, the cache doesn't save anything because the dependencies files are actually stored in a tempory folder in order to vendor them: https://github.com/perses/perses/blob/main/internal/cli/cmd/plugin/build/build.go#L286-L289

Which makes the whole caching system useless ...

andreasgerstmayr:
> in the `build` job, the cache doesn't save anything because the dependencies files are actually stored in a temporary folder in order to vendor them: https://github.com/perses/perses/blob/main/internal/cli/cmd/plugin/build/build.go#L286-L289
> 
> Which makes the whole caching system useless ...

Argh :(

Is there a way to prefetch the dependency? The only CUE dependency we need is the one we maintain ourself, shouldn't there be a way to use it without relying on the external CUE registry?

Nexucis:
@andreasgerstmayr thanks to https://github.com/perses/perses/pull/2795, I managed to fix the build of the plugin here using the cuelang cache.

Should be good to merge this one and see alive if it is indeed working from a fork :)

andreasgerstmayr (approved):

andreasgerstmayr:
nice work :100:

PR #91, opened 2025-03-18
[DOC] Fix some links & examples

AntoineThebaud:
- Fix some links & examples
- Also add mdox-based doc validation in CI to avoid missing such broken links next time.


Nexucis (approved):
awsome \o/

PR #101, opened 2025-03-24
[IGNORE] update dependencies

andreasgerstmayr:
update all frontend dependencies


Nexucis (commented):
really cool you did that @andreasgerstmayr.

Just a suggestion, on perses `immer` has been updated also. Perhaps you should do the same here. WDYT ? 

```
"immer": "^10.1.1",
```

andreasgerstmayr:
> really cool you did that @andreasgerstmayr.
> 
> Just a suggestion, on perses `immer` has been updated also. Perhaps you should do the same here. WDYT ?
> 
> ```
> "immer": "^10.1.1",
> ```

Yep :+1:, I updated the PR.
I think we need to automate syncing of the dependency versions in the future.

andreasgerstmayr:
looks like there were some breaking changes in `immer`. I'll fix it next week.

Nexucis:
yeah I guess you need to do the same changes than on https://github.com/perses/perses/pull/2774

Gladorme (approved):

Nexucis (approved):

PR #102, opened 2025-03-25
[ENHANCEMENT] Various schemas improvements

AntoineThebaud:
WIP, depends on new common libs to be available & released on perses/perses side first


Nexucis (approved):

jgbernalp:
Should we merge this before https://github.com/perses/plugins/pull/155? Then we can bump the version.

PR #103, opened 2025-03-28
[IGNORE] CI: use percli instead of custom script to validate schemas

AntoineThebaud:
- use percli instead of custom script to validate schemas
- new shared package for running commands in scripts


Nexucis (approved):
😍

.github/workflows/ci.yml
```diff
@@ -50,7 +50,7 @@ jobs:
             **/dist/mf-manifest.json
             !node_modules
 
-  lint:
+  lint-npm:
     name: 'lint'
```
Nexucis, line 54: you should change the name also

```suggestion
name: 'lint-npm'
```

PR #110, opened 2025-04-21
[FEATURE] Add DatasourceVariable plugin

OrReuben:
Connected to https://github.com/perses/perses/pull/2876
Closes https://github.com/perses/perses/issues/1514.

- Prometheus plugin - Supports the new value type from `DatasourceSelect`
- Tempo plugin - Supports the new value type from `DatasourceSelect`
- StaticListVariable plugin - Added a new list plugin type for `DatasourceVariable`:
   - Responsible of showcasing all available datasources based on selected kind.

Preview:
new

This pull request introduces enhancements to support variable-based datasource selection in Prometheus and Tempo plugins, along with the addition of a new `DatasourceVariable` plugin. Key changes include updates to the datasource handling logic, the introduction of new utility functions, and the implementation of the `DatasourceVariable` plugin to enable dynamic datasource selection.

### Enhancements to Prometheus and Tempo Plugins:

* Updated the `Prometheus` and `Tempo` plugins to support variable-based datasources by replacing `DatasourceSelector` with `DatasourceSelectValue` in their respective models and query editors. This allows for dynamic datasource selection using variables. [[1]](diffhunk://#diff-59f7392a66c2b8ac44c6bfe95bf24cd09c32802dc7cb031568be5e73c53981aeL33-R44) [[2]](diffhunk://#diff-d228d70511175f58ca34a0ce1612ed8889c4398e7a9c3d7da0594eaa8b765323L33-R44) [[3]](diffhunk://#diff-38754fc3fb511bae9143c45817f5d3d8776d2d511e1efbbecf00ac9de5972635L25-R26) [[4]](diffhunk://#diff-fec06ff07c632d8fc9b9a98b31238eae8bd531f9709c5d160df6d50cbd17c846R14-R22)
* Added utility functions such as `isVariableDatasource` and `datasourceSelectValueToSelector` to handle variable-based datasource logic. These functions are integrated into query execution and editor components. [[1]](diffhunk://#diff-a83cec3dd9b65a2c09dcff838d75b7560340b4e6b3b186c7d07c0301afab04d4L33-R34) [[2]](diffhunk://#diff-7a04d24c137b2e2a77978e7e210a3e1e48122f844ab841685e32b643af1c18e0R51-R61) [[3]](diffhunk://#diff-6ff35a3efa21e510eb00a73b9f94c763032b514873f95ba8682aa327e86905c6L56-R61)

### Addition of `DatasourceVariable` Plugin:

* Implemented a new `DatasourceVariable` plugin to dynamically populate datasource options based on the selected plugin kind. This includes a custom options editor for selecting datasource plugin kinds and fetching datasource options. [[1]](diffhunk://#diff-ad4443c666fdd0ed6e3a561b90c0814ce82d0ba95d333db76e98f9f5cbb03d6aR44-R52) [[2]](diffhunk://#diff-c80725925c960401aa1a0e0bbd71dbf7de7bb578d425ae07531680c4d54d9a35R39) [[3]](diffhunk://#diff-0b4da3ecd4818a695a6a44d450bdf8b0885a75a7de8d4e3cbcff604c488d70d7R1-R77)

### Refactoring and Code Simplification:

* Refactored query editors for Prometheus and Tempo to use the `useDatasourceSelectValueToSelector` hook for consistent handling of variable-based datasources. [[1]](diffhunk://#diff-65a28d02cb960c7f84eee2cd9933c018a77d461d1e978864c8f334d56dc8199cL42-R57) [[2]](diffhunk://#diff-9ce0753f10725e622b70b6ee6d70a61f8a9856fd7562e76669e3fd041296438bL32-R38)
* Updated query execution logic in Prometheus and Tempo to use the resolved datasource selector, ensuring compatibility with both static and variable-based datasources. [[1]](diffhunk://#diff-7a04d24c137b2e2a77978e7e210a3e1e48122f844ab841685e32b643af1c18e0L90-R100) [[2]](diffhunk://#diff-6ff35a3efa21e510eb00a73b9f94c763032b514873f95ba8682aa327e86905c6L56-R61)

These changes collectively improve the flexibility and usability of datasource selection in the system, enabling dynamic configurations through variables.


staticlistvariable/rsbuild.config.ts
```diff
@@ -36,6 +36,7 @@ export default defineConfig({
           name: 'StaticListVariable',
           exposes: {
             './StaticListVariable': './src/StaticListVariable.tsx',
+            './DatasourceVariable': './src/DatasourceVariable.tsx',
```
AntoineThebaud, line 39: Any particular reason you made this a part of the StaticListVariable plugin instead of making it a whole new plugin?

OrReuben: Sounds like the right place for it as it's a static list for datasources

AntoineThebaud: Well the current StaticListVariable is static in the sense that the list of values it holds is hardcoded in the variable definition. Here the list of values is dynamic, since adding or removing datasources will have an impact on it.
Anyway this isn't about being picky with naming: StaticListVariable and DatasourceVariable can be used independently and serve separate purposes, so someone might want to install one and not the other.

FYI, the ability to define multiple plugins within the same package was initially introduced for the case of datasource plugins, which are quite tightly coupled with their related variables (see for example [prometheus](https://github.com/perses/plugins/blob/main/prometheus/package.json#L46)). So it made sense to strongly couple them by making them part of the same package. We could imagine other use cases beyond this one, but here I don't think coupling StaticListVariable and DatasourceVariable makes enough sense.

Considering that, would you mind moving DatasourceVariable to a new plugin?

OrReuben: Makes sense, will move it to be a seperate plugin.

OrReuben: Also, just making sure - I remember from our chat we talked about it to be as one of the core plugins as it is not something one would ever want to uninstall. the StaticListVariable is a core plugin, so maybe it does make sense to make some kind of a plugin for core plugins

AntoineThebaud: That could be an option indeed, but it's still a bit opinionated to do so, and we have other plugins like TimeSeriesChart that are unlikely to be uninstalled I guess but still come as an independent package. Thus I'd prefer to have them separated for now, as if we were to do such grouping now or later we would anyway have to rename the top plugin name (currently StaticListVariable) to something else (CorePlugins?), causing a breaking change in all cases.

OrReuben: @AntoineThebaud separated to different plugin

AntoineThebaud:
Please also add the new folder to the list of workspaces in the main package.json ([link](https://github.com/perses/plugins/blob/main/package.json#L14)) - otherwise multiple CI jobs are blind to the new folder.

datasourcevariable/schemas/migrate/migrate.cue
```diff
@@ -0,0 +1,24 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+package migrate
+
+#grafanaType: "text"
```
AntoineThebaud, line 16: This line should be removed. `#grafanaType` works only in the case of panel plugins (https://github.com/perses/perses/blob/main/docs/plugins/cue.md)

datasourcevariable/package.json
```diff
@@ -0,0 +1,47 @@
...
+  "scripts": {
+    "dev": "rsbuild dev",
+    "build": "rsbuild build",
+    "lint": "eslint src --ext .ts,.tsx",
+    "test": "cross-env LC_ALL=C TZ=UTC jest --passWithNoTests",
+    "type-check": "tsc --noEmit"
+  },
+  "dependencies": {
+    "@module-federation/enhanced": "^0.8.9",
+    "color-hash": "^2.0.2"
```
jgbernalp, line 14:
```suggestion
```

jgbernalp:
@OrReuben Great addition, aside form the small dependency comment, LGTM. [Changes](https://github.com/perses/perses/pull/2876) from the Perses side also have a small comment but it should be good to go. Once the changes in Perses core and plugin system are merged, we can do a snapshot or beta release and update the dependencies here.

datasourcevariable/package.json
```diff
@@ -0,0 +1,47 @@
+{
+  "name": "@perses-dev/datasource-variable",
```
jgbernalp, line 2:
```suggestion
  "name": "@perses-dev/datasource-variable-plugin",
```

datasourcevariable/package.json
```diff
@@ -0,0 +1,47 @@
...
+  "private": true,
+  "version": "0.2.0",
```
jgbernalp, line 4:
```suggestion
  "version": "0.1.0",
```

AntoineThebaud (commented):
I guess it's fine now, most probably just need a `npm install` and a rebase

jgbernalp:
@OrReuben mind taking a look at the conflicts?

OrReuben:
@jgbernalp what is missing here to make it pass the pipeline?

jgbernalp:
it seems you have type errors, you can check them here: https://github.com/perses/plugins/actions/runs/14992572021/job/42130038662?pr=110

OrReuben:
@jgbernalp I'm pretty sure it happens because of we need to sync the plugins repo with the new released perses version.

The question if to update it here or do it in another PR

jgbernalp:
We should do it here, PTAL: https://github.com/logzio/perses-plugins/pull/1

OrReuben:
@jgbernalp thanks for the help! merged your PR

PR #128, opened 2025-05-09
[IGNORE] Migrate bar chart component

sjcobb:
Migrate BarChart, rename to BarChartBase
https://github.com/perses/perses/blob/v0.51.0-beta.1/ui/components/src/BarChart/BarChart.tsx


AntoineThebaud (approved):
🙏

PR #129, opened 2025-05-09
[IGNORE] Migrate pie chart component

sjcobb:
Migrate PieChart, rename to PieChartBase
https://github.com/perses/perses/blob/v0.51.0-beta.1/ui/components/src/PieChart/PieChart.tsx

### DIFF
--- a/piechart/src/PieChartPanel.tsx
+++ b/piechart/src/PieChartPanel.tsx
@@ -17,7 +17,6 @@ import {
   ContentWithLegend,
   LegendItem,
   LegendProps,
-  PieChart,
   PieChartData,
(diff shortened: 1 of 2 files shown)


piechart/src/PieChartPanel.tsx
```diff
@@ -17,7 +17,6 @@ import {
   ContentWithLegend,
   LegendItem,
   LegendProps,
-  PieChart,
   PieChartData,
```
AntoineThebaud, line 20: PieChartData import should be removed too

AntoineThebaud (approved):

PR #170, opened 2025-06-19
[BUGFIX] make Format field a pointer to keep optional

coleenquadros:
Importing JSON for table panel kind with empty format for the columns throws invalid panel error
```
              "columnSettings": [
                {
                  "name": "cluster",
                  "header": "Cluster",
                  "format": {
                    "unit": ""
                  },
                  "align": "right"
                },
```

`bad request: invalid panel 0_0: spec.columnSettings.0.format: 11 errors in empty disjunction: spec.columnSettings.0.format.unit: 8 errors in empty disjunction: spec.columnSettings.0.format.unit: conflicting values "bytes" and "": 1:77 ../plugins/Table-0.7.1/cue.mod/pkg/github.com/perses/perses/cue/common/format.cue:35:18 spec.columnSettings.0.format.unit: conflicting values "days" and "": 1:77 ../plugins/Table-0.7.1/cue.mod/pkg/github.com/perses/perses/cue/common/format.cue:19:69 spec.columnSettings.0.format.unit: conflicting values "decimal" and "": 1:77 ../plugins/Table-0.7.1/cue.mod/pkg/github.com/perses/perses/cue/common/format.cue:29:18 spec.columnSettings.0.format.unit: conflicting values "hours" and "": 1:77 ../plugins/Table-0.7.1/cue.mod/pkg/github.com/perses/perses/cue/common/format.cue:19:59 spec.columnSettings.0.format.unit: conflicting values "milliseconds" and "": 1:77 ../plugins/Table-0.7.1/cue.mod/pkg/github.com/perses/perses/cue/common/format.cue:19:18 spec.columnSettings.0.format.unit: conflicting values "minutes" and "": 1:77 ../plugins/Table-0.7.1/cue.mod/pkg/github.com/perses/perses/cue/common/format.cue:19:47 spec.columnSettings.0.format.unit: conflicting values "months" and "": 1:77 ../plugins/Table-0.7.1/cue.mod/pkg/github.com/perses/perses/cue/common/format.cue:19:88 spec.columnSettings.0.format.unit: conflicting values "seconds" and "": 1:77 ../plugins/Table-0.7.1/cue.mod/pkg/github.com/perses/perses/cue/common/format.cue:19:35 spec.columnSettings.0.format.unit: conflicting values "weeks" and "": 1:77 ../plugins/Table-0.7.1/cue.mod/pkg/github.com/perses/perses/cue/common/format.cue:19:78 spec.columnSettings.0.format.unit: conflicting values "years" and "": 1:77 ../plugins/Table-`

### DIFF
--- a/table/sdk/go/table.go
+++ b/table/sdk/go/table.go
@@ -47,16 +47,16 @@ const (
 )
 
 type ColumnSettings struct {
-	Name              string        `json:"name" yaml:"name"`
-	Header            string        `json:"header,omitempty" yaml:"header,omitempty"`
-	HeaderDescription string        `json:"headerDescription,omitempty" yaml:"headerDescription,omitempty"`
-	CellDescription   string        `json:"cellDescription,omitempty" yaml:"cellDescription,omitempty"`
-	Format            common.Format `json:"format,omitempty" yaml:"format,omitempty"`
-	Align             Align         `json:"align,omitempty" yaml:"align,omitempty"`
-	EnableSorting     bool          `json:"enableSorting,omitempty" yaml:"enableSorting,omitempty"`
-	Sort              Sort          `json:"sort,omitempty" yaml:"sort,omitempty"`
-	Width             float64       `json:"width,omitempty" yaml:"width,omitempty"`
-	Hide              bool          `json:"hide,omitempty" yaml:"hide,omitempty"`
+	Name              string         `json:"name" yaml:"name"`
+	Header            string         `json:"header,omitempty" yaml:"header,omitempty"`
+	HeaderDescription string         `json:"headerDescription,omitempty" yaml:"headerDescription,omitempty"`
+	CellDescription   string         `json:"cellDescription,omitempty" yaml:"cellDescription,omitempty"`
+	Format            *common.Format `json:"format,omitempty" yaml:"format,omitempty"`
+	Align             Align          `json:"align,omitempty" yaml:"align,omitempty"`
+	EnableSorting     bool           `json:"enableSorting,omitempty" yaml:"enableSorting,omitempty"`
+	Sort              Sort           `json:"sort,omitempty" yaml:"sort,omitempty"`
+	Width             float64        `json:"width,omitempty" yaml:"width,omitempty"`
+	Hide              bool           `json:"hide,omitempty" yaml:"hide,omitempty"`
 }
 
 type ValueConditionSpec struct {


saswatamcode:
Hmm actually I think the fix needs to be else where
Ideally, we can already leave format empty, and it would just be omitted from the generated config

The problem is, there is no omitempty on the unit field in common.Format, so it just sticks around, even if nothing was specified in the Go SDK. And the generated config then contains unit: "", which is really not a valid schema

So I think we need to add an omitempty to unit struct tag here. Wdyt @Nexucis? 🙂 

Nice catch @coleenquadros!

Nexucis:
the issue is more about Cuelang is considering an empty string as value. This issue will happen again and again everytime we have an empty JSON struct :/

AntoineThebaud:
So currently format for the table panel (others follow the same logic actually) is defined as such:

the `format` attribute itself is optional:
`format?: common.#format`

but if `format` is provided, then it's mandatory to provide the `unit` field within it & we expect a proper string value it, no emptry string allowed, e.g:
`unit: "milliseconds" | "seconds" | "minutes" | "hours" | "days" | "weeks" | "months" | "years"`

So personally I'd say that it makes no sense to support empty string for `unit`, the question is more about whether to make this field optional just like `format` is. So that's the final question here, does not providing a `unit` makes sense in the case you still wanted to provide a `format`? If I'm not mistaken I think the current philosophy was that, if you don't want to specify a precise unit (like bytes, minutes..) and are just interested into tuning the other `format` attributes ( `decimalPlaces` and sometimes as well `shortValues`) then you should actually rely on the generic unit `decimal`. This one can be considered a default/fallback option here and we are currently requiring people to explicitely set it in such case. Do we want to change that?

AntoineThebaud:
Personally I have no strong opinion on that so I'd be fine to make unit optional and thus to have the validation accepting
```json
"format": {},
```
But I wouldn't want that we accept this:
```json
"format": {
    "unit": ""
},
```

saswatamcode:
Yup, I would really prefer then,

Cue schema remains same
Go SDK uses omitempty for all optional things (since the idea is that we will end up generating YAML from it anyway, which will be applied and Go through CUE validation), and we skip pointer approach

Wdyt?

coleenquadros:
hi @saswatamcode are we okay to merge this fix to stay consistent with the other panels? This is still needed for table panels.

PR #173, opened 2025-06-23
[FEATURE] Add X axis exponential to the histogram

shahrokni:
# Description
Relates to #2977

This PR adds the X-Axis exponential feature on top of the standard Histogram chart.

## What is X-Axis Exponential?

X-Axis Exponential feature does not use the linear buckets passed down to the Histogram chart. Instead, based on the received linear buckets, exponential buckets are generated, and the data is redistributed into those exponential buckets. To generate exponential buckets a `base` should be given to the component. Although theoretically, base can be any arbitrary number, the here user should pick either 2, 4, 8, or 10. **_Later 0.5 will be added for Trigonometry applications. Likely with the next relevant PR._**

To clarify things, let's take a look at the following visual example.
The following data set shows the employees' salary range.

```javaScript
const incomeEmployees: HistogramChartData = {
        buckets: [
          [0, '500', '1000', '20'],
          [1, '1000', '1500', '200'],
          [2, '1500', '2000', '350'],
          [3, '2000', '2500', '300'],
          [4, '2500', '3000', '250'],
          [5, '3000', '3500', '200'],
          [6, '3500', '4000', '100'],
          [7, '4000', '4500', '20'],
          [8, '4500', '5000', '2'],
          [9, '5000', '5500', '5'],
        ],
      };
````

**With a linear bucket histogram:**

image

**With the exponential buckets:**

image

The second picture shows how the data is placed into exponential buckets which have been generated with base 2.
As you can see here the buckets are growing exponentially and getting bigger and bigger.

## Why do I see a +8000 bucket, while the max income is 5500?
Just take a look at the exponentials of 2 and see how it suddenly grows. 
Remember, Histogram only shows distribution under buckets. Here as you can see 5500 falls under the following bucket. 

```json
[    
    {
        "lb": 4096,
        "ub": 8192
    },
]
```

> To see a handful of exponential outputs with base 2, simply run the following script

```javaScript
const exponetialBuckets = [];
for (let i = 0; i < 20; i++) {
    const bucket = {
        lb: Math.pow(2, i),
        ub: Math.pow(2, i + 1)
    }
    exponetialBuckets.push(bucket);
}
console.log(exponetialBuckets);
```

## Why the X-Axis ends at 8692, while the exponential of 2 should be 8192? 

Because the income does not start from 0$. Here the minimum income is 500$. Therefore, the max is technically, 8692.

## Why do I see a different range on Y-Axis? 
Because, the buckets are growing, and they include more `counts`, This means, the Y-Axis is adjusting itself accordingly.

## What is the best base for my scenario?

Although there are techniques to calculate the best base to generate exponential buckets based on specific scenarios, at the moment we are restricted to 2, 4, 8 , and 10.  The next PR may expose a functionality which could calculate the best base accordingly. 

## Does this cover Trigonometry ranges (between 0 and 1) like  Sin(θ) and Cos(θ)?

**Short answer: Nope. I am working on it. Probably with the next relevant PR, it will be ready.**  ⚠️ 

Long answer: As you know there are infinite numbers between and 0-1. This means we can only get closer to 1, and never get there! This means, we need to somehow generate static buckets and cutoff the loop at some points. At, the moment I am working on this to make sure the functionality is reliable. ✅ ⌚ 

## How to test with my own dataset? 🧪 

There is a storybook file ready for you to test your own dataset. Simply put your data set there and run the following command

> npm run storybook

# Checklist

## UI Changes

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -12,14 +12,7 @@
 // limitations under the License.
 
 module.exports = {
-  extends: [
-    'eslint:recommended',
-    'plugin:react/recommended',
-    'plugin:react-hooks/recommended',
-    'plugin:jsx-a11y/recommended',
-    'plugin:@typescript-eslint/recommended',
-    'plugin:prettier/recommended',
-  ],
+  extends: ['eslint:recommended', 'plugin:react/recommended', 'plugin:react-hooks/recommended', 'plugin:jsx-a11y/recommended', 'plugin:@typescript-eslint/recommended', 'plugin:prettier/recommended', 'plugin:storybook/recommended'],
 
   plugins: ['import'],
 
--- a/.gitignore
+++ b/.gitignore
@@ -22,3 +22,9 @@ dist/
 
 # external CUE dependencies
 /*/cue.mod/pkg/
+
+*storybook.log
+storybook-static
+
+**/.storybook/
+**/stories/
--- a/histogramchart/package.json
+++ b/histogramchart/package.json
@@ -10,7 +10,9 @@
     "build:types": "tsc --project tsconfig.build.json",
     "lint": "eslint src --ext .ts,.tsx",
     "test": "cross-env LC_ALL=C TZ=UTC jest",
-    "type-check": "tsc --noEmit"
+    "type-check": "tsc --noEmit",
+    "storybook": "storybook dev -p 6006",
+    "build-storybook": "storybook build"
   },
   "main": "lib/cjs/index.js",
   "module": "lib/index.js",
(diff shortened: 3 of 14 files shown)


shahrokni:
> Thnx for this contribution @shahrokni. We probably need to adjust the cue model and the `HistogramChartOptionsEditorSettings` to allow users to toggle the exponential transformation.

Hello and thank you for the review,
I already added the settings to the  following files

1. histogramchart/src/histogram-chart-model.ts (Here I added the exponential setting)
2. histogramchart/src/components/HistogramChartOptionsEditorSettings.tsx (Here users can change value of the exponential base from None-10)
3. histogramchart/src/components/HistogramChartPanel.tsx (Here there exponential settings passed down to the chart)

Abut the cue  model you mentioned, I could not find such a thing for histogram chart. Maybe because it is purely front end? I am not sure :p ... 

@Gladorme

PR #206, opened 2025-07-28
[FEATURE] prometheus: Round the promql response time

shahrokni:
Closes https://github.com/perses/perses/issues/3133

# Description

As suggested here (https://github.com/perses/perses/issues/3133), the response time is simply rounded using `Math.round` function. 

# Test

To test this you need to run the Perses/Plugin localy

- Clone the Repo
- Swith to this branch
- Have the Perses CLI ready on your machine, and then:
-  ./percli.exe plugin start ../../../projects/plugins/prometheus/
- You also need to run the server and client of Perses itself

# Screenshots

image

# Checklist

## UI Changes

### DIFF
--- a/prometheus/src/components/TreeNode.tsx
+++ b/prometheus/src/components/TreeNode.tsx
@@ -383,7 +383,7 @@ function QueryStatus({
       <Typography variant="body2" component="span" sx={{ color: (theme) => theme.palette.grey[500] }}>
         {resultStats.numSeries} result{resultStats.numSeries !== 1 && 's'}
         &nbsp;&nbsp;–&nbsp;&nbsp;
-        {responseTime}ms
+        {responseTime ? `${Math.round(responseTime)}ms` : '? ms'}
         {resultStats.sortedLabelCards.length > 0 && <>&nbsp;&nbsp;–</>}
       </Typography>
       {resultStats.sortedLabelCards.slice(0, maxLabelNames).map(([ln, cnt]) => (


prometheus/src/components/TreeNode.tsx
```diff
@@ -383,7 +383,7 @@ function QueryStatus({
       <Typography variant="body2" component="span" sx={{ color: (theme) => theme.palette.grey[500] }}>
         {resultStats.numSeries} result{resultStats.numSeries !== 1 && 's'}
         &nbsp;&nbsp;–&nbsp;&nbsp;
-        {responseTime}ms
+        {Math.round(responseTime as number)}ms
```
Gladorme, line 386: If responseTime can be undefined, let handle this case too
```suggestion
        {responseTime ? `${Math.round(responseTime)}ms` : "Unknown response time"}
```

AntoineThebaud: I'd say "unknown" or "N/A" to keep it a short string on screen

Gladorme (approved):
lgtm

prometheus/src/components/TreeNode.tsx
AntoineThebaud: > N/A or empty string? Because N/A or unknown without context can be strange, no?

valid point, what about `? ms` or `— ms` then?

Gladorme: I prefer ? ms, else we will have: `@ requests - -ms - ...`

shahrokni: > I prefer ? ms, else we will have: `34 requests - - ms - ...`

Yes makes sense

AntoineThebaud (approved):
🙏

PR #210, opened 2025-07-29
[BREAKINGCHANGE] TracingGanttChart: configure attribute links in panel spec instead of panel props

andreasgerstmayr:
# Description

Previously, the attribute links (linking from span attributes to custom pages) were configured via panel props. However, this works only if the panel is embedded directly, i.e. `<TracingGanttChart.PanelComponent .../>` and doesn't work when included via `<Panel definition={{...}} />`.

This was a problem, because the TempoExplorer in the Tempo plugin includes the panel via the plugin system (via `<Panel ... />`) and therefore cannot pass custom props to the panel plugin.

This change moves the attribute link configuration to the panel spec.

# Screenshots

no UI changes

# Checklist

## UI Changes

### DIFF
--- a/tracingganttchart/src/TracingGanttChart/utils.ts
+++ b/tracingganttchart/src/TracingGanttChart/utils.ts
@@ -14,6 +14,7 @@
 import { PersesChartsTheme } from '@perses-dev/components';
 import { Theme } from '@mui/material';
 import { otlptracev1 } from '@perses-dev/core';
+import { replaceVariables, VariableStateMap } from '@perses-dev/plugin-system';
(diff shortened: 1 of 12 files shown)


tracingganttchart/src/TracingGanttChart/utils.ts
```diff
@@ -68,3 +69,15 @@ export function formatDuration(timeMs: number): string {
   }
   return `${+(timeMs / 1000).toFixed(2)}s`;
 }
+
+export function renderTemplate(
```
jgbernalp, line 73: do we already have this utility in the core or plugin system packages? it seems something we probably use already for queries

PR #213, opened 2025-07-30
[BUGFIX] Set the end of the time range according to the start and step

shahrokni:
Closes https://github.com/perses/perses/issues/2362

# Description

If the selected range of the time-series is shorter than the step, 
The `start` and `end` would be rounded to the same value. Consequently, the time range would be zero.
If the time range is set to zero, the X-Axis scale would not show the proper information. The query also would not be correct, because the range is zero.

This change set the end accordingly to make sure there would be always a valid time range.

image

# Checklist

## UI Changes


AntoineThebaud (approved):

AntoineThebaud:
Thanks for the fix!

Issue #92, opened 2025-03-18
Module federation warnings "No required version specified"


andreasgerstmayr:
There are a lot of warnings in the build:

```bash
/home/runner/work/plugins/plugins/node_modules/@perses-dev/components/dist/Legend/Legend.js:1:1
  ⚠ No required version specified and unable to automatically determine one. Unable to find required version for "immer" in description file/s
  │ /home/runner/work/plugins/plugins/package.json
  │ /home/runner/work/plugins/plugins/node_modules/@perses-dev/components/package.json
  │ It need to be in dependencies, devDependencies or peerDependencies. file: shared module immer
 @ ../node_modules/@perses-dev/components/dist/Legend/index.js
 @ ../node_modules/@perses-dev/components/dist/index.js
 @ consume shared module (default) @perses-dev/components@^0.51.0-beta.0 (strict) (fallback: /home/runner/work/plugins/plugins/node_modules/@perses-dev/components/dist/index.js)
 @ ./src/BarChartOptionsEditorSettings.tsx
 @ ./src/BarChart.ts
 @ container entry
/home/runner/work/plugins/plugins/node_modules/@perses-dev/components/dist/ThresholdsEditor/ThresholdsEditor.js:1:1
  ⚠ No required version specified and unable to automatically determine one. Unable to find required version for "immer" in description file/s
  │ /home/runner/work/plugins/plugins/node_modules/@perses-dev/components/package.json
  │ /home/runner/work/plugins/plugins/package.json
  │ It need to be in dependencies, devDependencies or peerDependencies. file: shared module immer
 @ ../node_modules/@perses-dev/components/dist/ThresholdsEditor/index.js
 @ ../node_modules/@perses-dev/components/dist/index.js
 @ consume shared module (default) @perses-dev/components@^0.51.0-beta.0 (strict) (fallback: /home/runner/work/plugins/plugins/node_modules/@perses-dev/components/dist/index.js)
 @ ./src/BarChartOptionsEditorSettings.tsx
 @ ./src/BarChart.ts
 @ container entry
/home/runner/work/plugins/plugins/node_modules/@perses-dev/plugin-system/dist/remote/PluginRuntime.js:1:1
  ⚠ No required version specified and unable to automatically determine one. Unable to find required version for "@emotion/styled" in description file/s
  │ /home/runner/work/plugins/plugins/node_modules/@perses-dev/plugin-system/package.json
  │ /home/runner/work/plugins/plugins/package.json
[...]

PR #215, opened 2025-07-31
[IGNORE] Move list of shared modules to rsbuild.shared.ts, fix build warnings

andreasgerstmayr:
# Description

Currently, every module declares its own shared modules and peer dependencies. Some plugins used shared modules but didn't declare them as shared, the versions of React were hardcoded.

This PR moves the list of shared modules to `rsbuild.shared.ts`, which will be a single place which should be kept in sync with the definitions of exposed shared modules in Perses core at https://github.com/perses/perses/blob/main/ui/plugin-system/src/remote/PluginRuntime.tsx.

I also consolidated all peer dependencies (i.e. all shared modules) in the root package, because imho it doesn't make sense that plugin A declares a peerDependency of the same package with a different version than plugin B, if all the plugins run in the same host application. Having them separate also makes bumping versions more cumbersome.

This also resolves the build warnings and warnings printed to the JS console described in #92.

# Screenshots

no UI changes

# Checklist

## UI Changes


Nexucis:
> Ah, I assumed the packages in the workspace inherit the dependencies of the root package.

that's the case when you are in the workspace. As soon as you are publishing the package, then it's not in the workspace anymore and it lost this inheritance sadly :(

PR #224, opened 2025-08-04
[BUGFIX] Fix statchar color for zero values

shahrokni:
@AntoineThebaud 

Closes https://github.com/perses/perses/issues/3142

## Bug Description 🐞 

Details of the bug can be found here https://github.com/perses/perses/issues/3142

## About the fix 🔧 

JS, the most lovely programming language in the world, equates `0` with `false`. Therefore, developers should be careful with their conditional statements. Since JS considers 0 to be false, if it makes sense for the given scenario, `typeof` should be used instead. 

For more information you can take a look at the following MDN link. 

https://developer.mozilla.org/en-US/docs/Glossary/Falsy

## Screenshots (After the fix)

image

# Checklist

## UI Changes

### DIFF
--- a/statchart/src/utils/get-color.ts
+++ b/statchart/src/utils/get-color.ts
@@ -27,7 +27,7 @@ export function getStatChartColor(
   // Determine the default color from thresholds or theme
   const defaultColor = thresholds?.defaultColor ?? chartsTheme.thresholds.defaultColor;
 
-  if (!value || (!thresholds?.steps && !mappings)) {
+  if (value === undefined || value === null || value === '' || (!thresholds?.steps && !mappings)) {
     return defaultColor;
   }
 
@@ -52,7 +52,7 @@ export function getStatChartColor(
 }
 
 function getColorFromMappings(value: StatChartValue, mappings: ValueMapping[]): string | null {
-  if (mappings?.length && value) {
+  if (mappings?.length && value !== undefined && value !== null) {
     const { color } = applyValueMapping(value, mappings);
     return color || null;
   }


ibakshay (approved):
🚀

statchart/src/utils/get-color.ts
```diff
@@ -52,7 +52,7 @@ export function getStatChartColor(
 }
 
 function getColorFromMappings(value: StatChartValue, mappings: ValueMapping[]): string | null {
-  if (mappings?.length && value) {
+  if (mappings?.length && typeof value === 'number') {
```
jgbernalp, line 55: applyValueMapping accepts also strings, this could break intended functionality.

```suggestion
  if (mappings?.length && typeof value !== null) {
```

statchart/src/utils/get-color.ts
```diff
@@ -27,7 +27,7 @@ export function getStatChartColor(
   // Determine the default color from thresholds or theme
   const defaultColor = thresholds?.defaultColor ?? chartsTheme.thresholds.defaultColor;
 
-  if (!value || (!thresholds?.steps && !mappings)) {
+  if (typeof value !== 'number' || (!thresholds?.steps && !mappings)) {
```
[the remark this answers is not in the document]
shahrokni: > We should also check for undefined I guess?

That is also true. Regarding the nature of the parameter, it can be undefined.

statchart/src/utils/get-color.ts
```diff
@@ -27,7 +27,12 @@ export function getStatChartColor(
   // Determine the default color from thresholds or theme
   const defaultColor = thresholds?.defaultColor ?? chartsTheme.thresholds.defaultColor;
 
-  if (!value || (!thresholds?.steps && !mappings)) {
+  if (
+    value === undefined ||
+    value === null ||
+    (typeof value === 'string' && !value) ||
```
AntoineThebaud, line 33: I guess this can be simplified to 
```suggestion
    value === '' ||
```

AntoineThebaud (approved):

PR #240, opened 2025-08-13
[IGNORE] Add migration logic test for most plugins

AntoineThebaud:
# Description

Add unit tests of migration schemas for most plugins. These tests come from the old big monolithic migration tests on perses/perses side - to be cleaned up with https://github.com/perses/perses/pull/3189.

# Checklist

### DIFF
--- a/prometheus/schemas/prometheus-promql/migrate/tests/no-aggr/expected.json
+++ b/prometheus/schemas/prometheus-promql/migrate/tests/no-aggr/expected.json
@@ -1,7 +1,7 @@
 {
   "kind": "PrometheusPromQLVariable",
   "spec": {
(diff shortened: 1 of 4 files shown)


barchart/schemas/migrate/tests/basic/expected.json
```diff
@@ -0,0 +1,10 @@
+{
+  "kind": "BarChart",
+  "spec": {
+    "calculation": "last-number",
+    "format": {
+      "decimalPlaces": 2,
+      "unit": "percent"
+    }
```
Nexucis, line 8: Maybe you didn't finish your work, but here I find it is strange that the query is not migrated and the datasource setup as well

PR #266, opened 2025-08-21
[IGNORE] fix build configuration for plugins

jgbernalp:
# Description

- fix configuration for building plugins, make older configuration consistent to use the new `@module-federation/rsbuild-plugin` package
- add index-federation.ts to the plugins that missed it
- Fixes the validation in `pyroscope/src/plugins/pyroscope-profile-query/get-profile-data.ts` this was not reported before as the typescript configuration was different.

# Checklist

### DIFF
--- a/flamechart/package.json
+++ b/flamechart/package.json
@@ -1,6 +1,14 @@
 {
-  "name": "@perses-dev/flame-chart",
+  "name": "@perses-dev/flame-chart-plugin",
   "version": "0.1.0",
+  "homepage": "https://github.com/perses/plugins/blob/main/README.md",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/perses/plugins.git"
+  },
+  "bugs": {
+    "url": "https://github.com/perses/plugins/issues"
+  },
   "scripts": {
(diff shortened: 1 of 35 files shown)


Gladorme (approved):
LGTM :)

flamechart/package.json
```diff
@@ -1,6 +1,14 @@
 {
   "name": "@perses-dev/flame-chart",
```
Nexucis, line 2: should we change the name also to suffix it by `-plugin` ?

Nexucis: and I think that's the same for the pyroscope plugin it is not suffix by `-plugin`

jgbernalp: yes, good catch. This will create new npm packages but I don't think that is an issue

Nexucis: yeah we didn't say anything about these new plugins, so we can still break them hehe

jgbernalp:
> by the way do you need to update the command `percli plugin generate` @jgbernalp with these changes ?

I was checking to update them, but the generation command already has the updates on the template.

Nexucis (approved):

PR #267, opened 2025-08-22
[ignore] add a script to bump perses-dev dependencies for all plugins

Nexucis:


scripts/bump-deps/bump-deps.go
```diff
@@ -0,0 +1,80 @@
...
+
+var bumpNPMDeps = regexp.MustCompile(`"@perses-dev/([a-zA-Z-]+)":\s*"(\^)?[0-9]+\.[0-9]+\.[0-9]+(-(alpha|beta|rc)\.[0-9]+)?"`)
+
+func bumpGoDep(workspace, version string) {
+	goGetCMD := command.Create("go", "get", fmt.Sprintf("github.com/perses/perses@v%s", version))
+	goGetCMD.Dir = workspace
+	if err := goGetCMD.Run(); err != nil {
+		logrus.WithError(err).WithField("workspace", workspace).Error("unable to bump the go dependencies")
+		return
+	}
```
shahrokni, line 35: Question, if we encounter any issue here, we log and return, and the for loop in the main continues, meaning we continue the process. Wouldn't this end up in a partial bump? Shouldn't we stop or redo the entire process?

shahrokni (approved):

PR #280, opened 2025-08-28
[ignore] generate changelog for plugin release

Nexucis:


scripts/release/changelog.go
```diff
@@ -0,0 +1,55 @@
...
+	if previousTag == "" {
+		logrus.Infof("no previous tag found for plugin %s, skipping changelog generation", pluginName)
+		return "First release"
+	}
+	logrus.Infof("previous tag for plugin %s is %s", pluginName, previousTag)
+	entries := changelog.GetGitLogs(previousTag)
+	var newEntries []string
+	for _, entry := range entries {
+		pluginName = strings.ToLower(pluginName)
+		if !strings.Contains(strings.ToLower(entry), pluginName) {
```
andreasgerstmayr, line 49: Does this check if the plugin name in lowercase is part of the commit message?

An alternative way to filter `git log` would be to run `git log <path_of_plugin>`, which filters the git log for any commits which contain changes in the specified path

andreasgerstmayr: > For example the re-formating files. Example: #212. The tempo files has been changed just because of a formating but it has nothing to do with the purpose of the PR

In this PR I updated the spec of the ScatterPlot, and I had to do this change also for the ScatterPlot panel definition in the Tempo explore page.

I guess in the end with the string search method we might miss some things (i.e. the plugin name must be exactly in the log message, not `Scatter Chart` or `Scatter Plot` or some slight variation), and with the git log method we may add too much information to the changelog.

I'm fine either way, my comment shouldn't block this PR.

andreasgerstmayr (approved):

scripts/release/changelog.go
AntoineThebaud: We can try to make it as smart as possible but we can't escape the fact that we should avoid (even more with an auto generated changelog than before) packing unrelated changes to different plugins in a same PR. Besides we sometimes have PRs applying the same kind of update to multiple plugins, and to easily support that with the auto changelog I think Andreas's proposal to rely on `git log -- <path_of_plugin>` is the way to go. Otherwise it means we should split this kind of PR into N parts, 1 for each plugin, which could quickly become tedious.

Having the plugin name as a prefix in the title is a good practice in any case imho (and maybe like "multiple", "wide" or no prefix at all for PRs affecting several plugins?). We could make it [an item in the PR template like for the catalog entry](https://github.com/perses/plugins/blob/main/.github/pull_request_template.md?plain=1#L17-L18).

PR #380, opened 2025-09-16
[ENHANCEMENT]: TimeSeriesChart Migration Fix

ibakshay:
## Problem
CUE validation error during Grafana to Perses migration:
```
invalid panel 1: spec.yAxis.max: cannot reference optional field: min
```

## Root Cause
The TimeSeriesChart migration script could set `yAxis.max` without `yAxis.min`, but the schema constraint `max?: number & >= min` requires both fields to be present for validation.

https://github.com/perses/plugins/blob/02f28cd27a3870e9c28186a9ee3ec145a184c7da/timeserieschart/schemas/time-series.cue#L55

## Solution
Modified `timeserieschart/schemas/migrate/migrate.cue` to handle yAxis min/max values together:

- When `max` is specified without `min`, automatically provide `min: 0` as a sensible default
- Ensures schema constraints are always satisfied during migration
- Added tests for this use-case

## Impact
- Fixes migration failures for Grafana panels with only max values set
- Prevents schema validation errors during dashboard migration

### DIFF
--- a/timeserieschart/schemas/migrate/migrate.cue
+++ b/timeserieschart/schemas/migrate/migrate.cue
@@ -81,9 +81,6 @@ spec: {
 		},
 		null,
 	][0]
-	if #min != null {
-		yAxis: min: #min
-	}
 
 	#max: [// switch
 		if (*#panel.fieldConfig.defaults.max | null) != null {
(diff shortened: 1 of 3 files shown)


AntoineThebaud (commented):
Well there's actually an issue in the datamodel here, not really in the migration schema, I must admit I didnt think about that case..

IMHO the model should be changed to:

```cue
min?:    number
max?:    number
if min != _|_ && max != _|_ {
    max: >= min
}
```

This way you can just provide the max & that works (https://cuelang.org/play/?id=BxZsOWQhd5m#w=function&i=cue&f=export&o=yaml)

ibakshay:
> Well there's actually an issue in the datamodel here, not really in the migration schema, I must admit I didnt think about that case..
> 
> IMHO the model should be changed to:
> 
> ```cue
> min?:    number
> max?:    number
> if min != _|_ && max != _|_ {
>     max: >= min
> }
> ```
> 
> This way you can just provide the max & that works (https://cuelang.org/play/?id=BxZsOWQhd5m#w=function&i=cue&f=export&o=yaml)

Thank you very much for the suggestion. 
I have fixed in https://github.com/perses/plugins/pull/380/commits/4ee343d0c7be1dedc798ea3e769719ecb16343e2

AntoineThebaud (approved):

PR #386, opened 2025-09-19
[FEATURE] TimeSeriesChart: support opacity & line style overrides

AntoineThebaud:
# Description

Closes https://github.com/perses/perses/issues/3227.
Closes https://github.com/perses/perses/issues/3356.

This PR moves the Query Settings of the TimeSeriesChart to a dedicated tab & enrich it with new capabilities to override line style & opacity for individual trends.

NB: migration logic for the equivalent Grafana features will be done in a follow-up PR.

# Screenshots

image

image

image

# Checklist

## UI Changes

### DIFF
--- a/timeserieschart/src/QuerySettingsEditor.tsx
+++ b/timeserieschart/src/QuerySettingsEditor.tsx
@@ -144,6 +144,20 @@ export function QuerySettingsEditor(props: TimeSeriesChartOptionsEditorProps): R
     }
   };
 
+  // Helper function to update query settings at a specific index
+  const updateQuerySettings = (i: number, updater: (qs: QuerySettingsOptions) => void): void => {
+    if (querySettingsList !== undefined) {
+      handleQuerySettingsChange(
+        produce(querySettingsList, (draft) => {
+          const qs = draft[i];
+          if (qs) {
+            updater(qs);
+          }
+        })
+      );
+    }
+  };
+
   const deleteQuerySettingsInput = (i: number): void => {
     if (querySettingsList !== undefined) {
       const updatedQuerySettingsList = produce(querySettingsList, (draft) => {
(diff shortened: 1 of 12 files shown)


timeserieschart/src/utils/data-transform.ts
```diff
@@ -115,11 +116,10 @@ export function getTimeSeries(
     symbolSize: pointRadius,
     lineStyle: {
       width: lineWidth,
-      opacity: 0.95,
-      type: visual.lineStyle,
+      type: (querySettings?.lineStyle ?? visual.lineStyle) as 'solid' | 'dashed' | 'dotted',
     },
```
shahrokni, line 120: Just an opinion. 
The lineStyle > type could have had its own dedicated type. 

```typeScript
type lineStyleType = 'dotted' | 'dashed' | 'solid';
```

timeserieschart/src/QuerySettingsEditor.tsx
```diff
@@ -82,12 +113,123 @@ export function QuerySettingsEditor({ querySettingsList, onChange }: QuerySettin
...
+  const removeAreaOpacity = (i: number): void => {
+    if (querySettingsList !== undefined) {
+      handleQuerySettingsChange(
+        produce(querySettingsList, (draft) => {
+          const qs = draft[i];
+          if (qs) {
+            qs.areaOpacity = undefined;
+          }
+        })
+      );
```
shahrokni, line 232: If I am not mistaken, apart from the part which sets the qs, the methods are almost identical. 

They check to assure  querySettingsList is not undefined, they get the  draft[i] and put into the qs, and finally,
Set the qs['X']  accordingly. Couldn't we have a single method to complete basic check and manage the rest with a switch-case?

timeserieschart/src/QuerySettingsEditor.tsx
```diff
@@ -125,19 +265,14 @@ export function QuerySettingsEditor({ querySettingsList, onChange }: QuerySettin
...
-    >
-      {querySettingsList && querySettingsList.length > 0 ? (
+    <Stack>
+      {queryCount === 0 ? (
+        <Typography mb={2} fontStyle="italic">
+          No query defined
+        </Typography>
+      ) : (
+        querySettingsList &&
+        querySettingsList.length > 0 &&
```
shahrokni, line 275: This could have been one single line?

1. The `?` causes a short_circuit  and turns to a falsy value
2. Any length greater than 0 is truthy value. 

```typescript
querySettingsList?.length &&        
```

timeserieschart/src/QuerySettingsEditor.tsx
```diff
@@ -149,66 +284,262 @@ export function QuerySettingsEditor({ querySettingsList, onChange }: QuerySettin
...
+    if (!colorMode) options.push({ key: 'color', label: 'Color', action: onAddColor });
+    if (!lineStyle) options.push({ key: 'lineStyle', label: 'Line Style', action: onAddLineStyle });
+    if (areaOpacity === undefined) options.push({ key: 'opacity', label: 'Opacity', action: onAddAreaOpacity });
+    return options;
+  }, [colorMode, lineStyle, areaOpacity, onAddColor, onAddLineStyle, onAddAreaOpacity]);
+
+  const handleAddMenuClick = (event: React.MouseEvent<HTMLElement>) => {
+    if (availableOptions.length === 1 && availableOptions[0]) {
+      // If only one option left, add it directly
+      availableOptions[0].action();
```
shahrokni, line 364: It seems we have an extra check here. If `availableOptions.length` is greater than 0 and true, it means there is already at least a member placed in the 0 index. Therefore, this check could simply be (unless you are mutating the array members somewhere else, which means the 0 index may be removed manually)

```typescript
if (availableOptions.length) {
   availableOptions[0].action();
}
```

AntoineThebaud: Hmm we have to ensure there is only one element for this check (not more, not less), but you're right that the second check is kinda redundant here, thus we can go with simply `if (availableOptions.length === 1)`

AntoineThebaud: actually the compiler doesnt like it :( Object is possibly undefined)
image

I'll the double check then

PR #394, opened 2025-09-30
[ENHANCEMENT] add getPublicPath function to all plugins

lachieh:
# Description

This PR implements [the changes that were suggested](https://github.com/perses/perses/pull/3377#issuecomment-3317888993) by @jgbernalp suggested that the publicPath in the plugins can be defined through the module federation configuration.

Follow on from #391

### DIFF
--- a/barchart/rsbuild.config.ts
+++ b/barchart/rsbuild.config.ts
@@ -56,6 +56,7 @@ export default defineConfig({
       },
       dts: false,
       runtime: false,
+      getPublicPath: `function() { const prefix = window.PERSES_PLUGIN_ASSETS_PATH || window.PERSES_APP_CONFIG?.api_prefix || ""; return prefix + "${assetPrefix}"; }`,
     }),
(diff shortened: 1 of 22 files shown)


barchart/rsbuild.config.ts
```diff
@@ -56,6 +56,7 @@ export default defineConfig({
       },
       dts: false,
       runtime: false,
+      getPublicPath: `function() { const prefix = window.PERSES_PLUGIN_ASSETS_PATH || window.PERSES_APP_CONFIG.api_prefix || ""; return prefix + "${assetPrefix}"; }`,
```
jgbernalp, line 59: I updated the comment with a simplified version that accounts for missing app config:

```
getPublicPath: `function() { return (window.PERSES_PLUGIN_ASSETS_PATH || window.PERSES_APP_CONFIG?.api_prefix || "") + "${assetPrefix}"; }`,
```

PR #485, opened 2025-11-24
[FEATURE] Prometheus: support query params

AntoineThebaud:
# Description

Closes https://github.com/perses/perses/issues/1523.
Closes https://github.com/perses/perses/issues/3509 (`engine` query param can be passed).

I guess this piece of code could be moved to plugin-system in the future if other datasources start to support query params. Since it's Prometheus only for now I prefer to have it here.

# Screenshots

New section at the end of the datasource editor:

Screenshot 2025-11-24 164405

# Checklist

## UI Changes

### DIFF
--- a/prometheus/src/plugins/PrometheusDatasourceEditor.tsx
+++ b/prometheus/src/plugins/PrometheusDatasourceEditor.tsx
@@ -16,10 +16,12 @@ import { HTTPSettingsEditor } from '@perses-dev/plugin-system';
 import { Box, TextField, Typography, IconButton } from '@mui/material';
 import PlusIcon from 'mdi-material-ui/Plus';
 import MinusIcon from 'mdi-material-ui/Minus';
-import React, { ReactElement, useState } from 'react';
+import React, { ReactElement, useState, useRef } from 'react';
 import { DEFAULT_SCRAPE_INTERVAL, PrometheusDatasourceSpec } from './types';
 
 interface QueryParamEntry {
+  // Unique identifier for the entry, added to avoid using array index as key
+  id: string;
   key: string;
   value: string;
 }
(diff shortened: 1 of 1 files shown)


prometheus/src/plugins/PrometheusDatasourceEditor.tsx
```diff
@@ -26,6 +33,69 @@ export interface PrometheusDatasourceEditorProps {
...
+  const hasDuplicates = duplicateKeys.size > 0;
+
+  // Convert entries array to object and trigger onChange
+  const syncToParent = (newEntries: QueryParamEntry[]) => {
+    const newParams: Record<string, string> = {};
+    newEntries.forEach(({ key, value }) => {
+      if (key !== '' || value !== '') {
+        newParams[key] = value;
+      }
+    });
```
shahrokni: ## Question
Let's say
1. key is empty
2. value is not empty
3. Now based on the condition
4. Does that mean newParams[''] = value?

prometheus/src/plugins/PrometheusDatasourceEditor.tsx
```diff
@@ -93,6 +165,57 @@ export function PrometheusDatasourceEditor(props: PrometheusDatasourceEditorProp
...
+          {entries.map((entry, index) => (
+            <Box key={index} display="flex" alignItems="center" gap={2} mb={1}>
+              <TextField
+                size="small"
+                label="Key"
+                value={entry.key}
+                placeholder="Parameter name"
+                disabled={isReadonly}
+                onChange={(e) => handleQueryParamChange(index, 'key', e.target.value)}
+                error={entry.key !== '' && duplicateKeys.has(entry.key)}
```
shahrokni: If the key is empty, it can not be found in the Set. Therefore, we should be able to safely remove the first condition.

```typescript
error={duplicateKeys.has(entry.key)}
```

prometheus/src/plugins/PrometheusDatasourceEditor.tsx
```diff
@@ -26,6 +33,69 @@ export interface PrometheusDatasourceEditorProps {
...
+    }
+    setEntries(newEntries);
+    syncToParent(newEntries);
+  };
+
+  const addQueryParam = () => {
+    const newEntries = [...entries, { key: '', value: '' }];
+    setEntries(newEntries);
+    syncToParent(newEntries);
+  };
```
shahrokni: # Let me test  this manually 🧪 

## Scenario 🖊️ 

1. Add a Query Param
2. Don't set the key and value
3. Try to apply and save
4. Check if you could persisted the payload

prometheus/src/plugins/PrometheusDatasourceEditor.tsx
AntoineThebaud: that's a miss indeed, empty key should be forbidden

PR #533, opened 2026-01-21
[BUGFIX] Logstable: all quries results must be included in the logs table

shahrokni:
# Description 🖊️ 
Working on https://github.com/perses/perses/issues/3786 I found a bug which may also lead to other bugs related to transform. Let's see!

Logs table should include the result of all queries. It seems the buggy version only includes the query result at index 0. 

image

# Checklist

## UI Changes


logstable/src/test/mock-query-results.ts
rickardsjp: This is good (reversing the timestamps) but only tests sorting and does not test whether results from multiple queries are included.

PR #545, opened 2026-01-28
[FEATURE] Table: migrate column datalink overrides

PeterYurkovich:
Signed-off-by: PeterYurkovich

# Description

This PR looks to enable migrating column specific data links from overrides. The perses column data links only currently supports a single link, so the last link is taken from the list of links that are added in the overrides. This PR does not take into account default data links set for the whole table.

# Checklist

### DIFF
--- a/table/schemas/migrate/tests/link/expected.json
+++ b/table/schemas/migrate/tests/link/expected.json
@@ -13,7 +13,7 @@
       {
         "name": "job",
         "dataLink": {
-          "url": "/d/job-dashboard?job=${__data.fields.job}",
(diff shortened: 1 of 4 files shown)


shahrokni:
This comment is beyond this review, and 99% is wrong. However, I could not resist to put a comment and ask the following question.

In the description, you have mentioned that 

> This PR does not take into account default data links set for the whole table.

I would like to mention that in Perses the entire table can be linked. So, the question would be couldn't the default data link be used for the table link? Does that make sense?
@AntoineThebaud 
@PeterYurkovich

PR #573, opened 2026-02-17
[FEATURE]: add CI e2e setup to plugins

shahrokni:
Relates to https://github.com/perses/perses/issues/3799

## E2E setup
This change adds e2e set up into the plugin repo. Using the `PERSES_PLUGIN_ARCHIVE_PATHS` parameter we are able to inject the built plugins (archive files) into the running instance of Perses.  For more information please take a look at https://perses.dev/perses/docs/configuration/configuration/

⚠️  Use a recent tag. The mentioned config is pretty new!  I used the following tag,

> main-2026-02-26-739dc289-distroless-debug

> Check: https://hub.docker.com/r/persesdev/perses

## How to test this

- Edit a plugin 
- For example you can change its name from rsbuild.config
- And just push
- You can check the plugins endpoint and see that the substitute archive is in place
- You can use the following code to assure your change is also available from the tests

```typescript
/* WAS ONLY ADDED FOR DEMONSTRATIONS*/
test.describe('table Plugin', () => {
  test('ONLY FOR PR DEMONSTRATION', async ({ request }) => {
    const res = await request.get('http://localhost:8080/api/v1/plugins');
    expect(res.ok()).toBeTruthy();
    const plugins = await res.json();
    console.log('--- Current Perses Plugins ---');
    console.log(
      JSON.stringify(
        Array.from(plugins).map((p) => p.spec),
        null,
        2
      )
    );
  });
});
```

> http://localhost:8080/api/v1/plugins 

image

## Provisioning 

Provisioning was also set up to ease the testing process. **_The content doesn't matter at this point. The aim was only to make sure it is ready for the further steps._** 

image

# Checklist

## UI Changes

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,6 +76,42 @@ jobs:
       - run: npm ci
       - run: npm run test
 
+  e2e:
+    name: "e2e tests"
+    needs: build # <--- CRITICAL: Wait for build to finish. PERSES_PLUGIN_ARCHIVE_PATHS accepts archive files only
+    runs-on: ubuntu-latest
+    steps:
+      - name: checkout
+        uses: actions/checkout@v6
+      - name: download plugin archives
+        uses: actions/download-artifact@v4 # <--- Download the artifacts to use the archive files
+        with:
+          name: archives
+          path: ./plugins/temp-archives
+      - name: flatten plugin archives # <--- Move all plugins archive into a single folder for simplicity
+        run: |
+          mkdir -p ./plugins/development-archives          
+          find ./plugins/temp-archives -name "*.tar.gz" -exec mv {} ./plugins/development-archives/ \;          
+          rm -rf ./plugins/temp-archives          
+          ls -R ./plugins/development-archives
+      - name: start dev env
+        uses: hoverkraft-tech/compose-action@v2.5.0
+        with:
+            compose-file: .github/workflows/docker-compose.ci.yml      
+      - uses: perses/github-actions@v0.11.0
+      - uses: ./.github/perses-ci/actions/setup_environment
+        with:
+          enable_npm: true
+          nvmrc_path: ".nvmrc"
+      - name: install dependencies
+        run: npm ci
+      - name: install Playwright browsers
+        run: npx playwright install --with-deps
+        working-directory: e2e
+      - name: run e2e tests
+        run: npm run test:ci --prefix e2e -- src/tests
+          
+
   type-check:
     name: "type-check"
     runs-on: ubuntu-latest
--- a/e2e/package.json
+++ b/e2e/package.json
@@ -1,0 +1,16 @@
+{
+  "name": "@perses-dev/plugins-e2e",
+  "version": "0.1.0",
+  "description": "e2e tests for plugins",
+  "private": true,
+  "scripts": {
+    "test:ci": "npx playwright test --workers=1 --reporter=github,list",
+    "test:install": "npx playwright install --with-deps",
+    "test:local": "npx playwright test --ui",
+    "build": "echo 'Skipping build for e2e tests'"
+  },
+  "devDependencies": {
+    "@playwright/test": "^1.58.2",
+    "@types/node": "^25.2.3"
+  }
+}
--- a/e2e/playwright.config.ts
+++ b/e2e/playwright.config.ts
@@ -1,0 +1,39 @@
+// Copyright The Perses Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { defineConfig, devices } from '@playwright/test';
+
+export default defineConfig({
+  testDir: './src/tests',
+  fullyParallel: true,
+  retries: 0,
+  reporter: 'list',
+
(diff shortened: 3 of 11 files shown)


.github/workflows/ci.yml
```diff
@@ -76,6 +76,75 @@ jobs:
...
+        with:      
+          node-version: 20
+          cache: npm
+      
+      - name: Install dependencies
+        run: |
+          npm ci
+          npm ci --prefix e2e
+      
+      - name: Install percli
```
Nexucis: we also have a GitHub action to install percli

```suggestion
      - name: install percli
        uses: perses/cli-actions/actions/install_percli@v0.2.0
        with:
          cli-version: "v0.53.0-rc.1"
```

Nexucis (commented):
overall, thank you for putting this in place.

I have added few suggestions that will improve the CI/CD and simplify it.

I will probably add more comments later, but this is already a good step

.github/workflows/ci.yml
```diff
@@ -76,6 +76,75 @@ jobs:
...
+      - name: Setup node
+        uses: actions/setup-node@v4
+        with:      
+          node-version: 20
+          cache: npm
+      
+      - name: Install dependencies
+        run: |
+          npm ci
+          npm ci --prefix e2e
```
Nexucis: I don't understand why do we need to install twice the dependencies. This is a monorepo, all dependencies should be installed once for ever

.github/workflows/ci.yml
```diff
@@ -76,6 +76,75 @@ jobs:
...
+          
+          # It is better to keep it a static array
+          PLUGINS=("table" "tracetable")
+
+          for plugin in "${PLUGINS[@]}"; do
+            cd "$plugin" && npm run dev -- --port 3005 > /dev/null 2>&1 &                    
+
+            sleep 10
+
+            percli plugin start "./$plugin" http://172.17.0.1:3005
```
Nexucis: I am pretty sure this line won't let you to continue as it won't give back the hand.

e2e/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "e2e",
```
Nexucis: this package should be part of the npm workspace

```suggestion
  "name": "@perses-dev/plugins-e2e",
  "version": "0.1.0",
  "description": "e2e tests for plugins",
  "private": true,
```

e2e/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "version": "1.0.0",
+  "main": "index.js",
+  "scripts": {
+    "test:ci": "npx playwright test --workers=1 --reporter=github,list",
+    "test:install": "npx playwright install --with-deps",
+    "test:local": "npx playwright test --ui"
+  },
+  "keywords": [],
+  "author": "",
+  "license": "ISC",
```
Nexucis: no need to define the license

e2e/package.json
```diff
@@ -0,0 +1,18 @@
+{
+  "name": "e2e",
+  "version": "1.0.0",
+  "main": "index.js",
```
Nexucis: there is no main entrypoint for this package

scripts/tidy-modules/tidy-modules.go
```diff
@@ -42,6 +42,12 @@ func tidyGoModule(workspace string) error {
 
 func main() {
 	for _, workspace := range npm.MustGetWorkspaces(".") {
+
+		if workspace == "e2e" {
```
Nexucis: This thing should be added everywhere you have a loop on the workspace, which happen on other golang script than the one already modified here.

.github/workflows/ci.yml
```diff
@@ -76,6 +76,46 @@ jobs:
...
+        run: npx playwright install --with-deps
+        working-directory: e2e        
+      - name: setup Perses and start Plugins
+        run: |
+          # Handshake with server on port 8080
+          percli login http://localhost:8080
+          PLUGINS=("table" "tracetable")
+          for plugin in "${PLUGINS[@]}"; do
+            percli plugin start "./$plugin" &
+            sleep 10
```
jgbernalp: This is probably not reliable and might create unnecessary delays. The setup environment can include go and we can move this to a go script that will also handle background process termination gracefully.

e2e/playwright.config.ts
```diff
@@ -0,0 +1,40 @@
...
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { defineConfig, devices } from '@playwright/test';
+
+export default defineConfig({
+  testDir: './tests',
+  fullyParallel: true,
+  retries: 0,
+  workers: 1,
```
jgbernalp: fullyParallel does not make sense with only 1 worker. We could leave the workers undefined so playwright selects based on available resources

scripts/manage_plugins/manage_plugins.go
```diff
@@ -0,0 +1,81 @@
...
+			cmd.Stderr = os.Stderr
+
+			logrus.Infof("Starting plugin: %s\n", p)
+			if err := cmd.Start(); err != nil {
+				logrus.Errorf("failed to start %s: %v\n", p, err)
+				return
+			}
+
+			// Instead of a blind sleep 10, could we do a health logic here? I am not sure! Ask Gab
+			// Hey Gab, Plugins start on an arbitrary port, how could we check if they are ruining or not?!
```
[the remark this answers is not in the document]
shahrokni: I used taskHelper as suggested to manage async tasks. OK, that is a good suggestion.

**However**, I would like to mention that the effort here is to eliminate the `time.sleep`  **between** **

- **Running a plugin** (on an **arbitrary** port using percli)
- **Running its relevant tests** (**when they are attached and ready**)

Simply put, we should wait for the following results, before we run any e2e tests

image


So the question is, when those plugins are running on arbitrary ports hidden from our script, how could we assure that the plugin is ready? If the ports were predictable, there could be a health check mechanism. 


Using taskHelper  alone, although a good suggestion, does not resolve the main issue here. 
If there is anything that I am missing, please let me know

PR #577, opened 2026-02-19
[FEATURE] Add LogExplorer plugin

rickardsjp:
# Description
  
This PR adds a Log Explorer for the growing number of log query datasources. The LogExplorer consists of a MultiQueryEditor for building log queries, an automatically-generated volume histogram showing log volume over time, and a LogsTable displaying the log entries. The histogram requires the log data source plugins to implement an optional `createVolumeQuery` interface method. This is currently implemented for Loki and VictoriaLogs, but not yet for ClickHouse.

#### VictoriaLogs
- Added optional `createVolumeQuery` method to its LogQueryPlugin interface
- Implemented with dynamic interval selection (targets ~40 histogram bars across time range)
- Generates `| stats by (_time:<interval>, _stream) count() as volume` queries
- Filters `__name__` metadata label from time series results for cleaner histogram legends

#### Loki  
- Added optional `createVolumeQuery` method to its LogQueryPlugin interface
- Implemented with dynamic interval selection
- Generates `sum by (level, detected_level) (count_over_time(...))` queries

#### ClickHouse
- Added optional `createVolumeQuery` method to its LogQueryPlugin interface
- Stub implementation returns `null` (requires SQL query parsing for proper implementation)

### Testing this PR:

Because this PR changes multiple plugins at once, you will need to build the plugins using `percli` and copy them to the `plugin_archive` directory of your perses server. I believe the standard plugin development server can only run one plugin at once. If you clone the victorialogs repo, you can start a sample instance with `make docker-vl-single-up`. For Loki, you can use [Grafana's instructions](https://grafana.com/docs/loki/latest/get-started/quick-start/quick-start/) to set up a Loki datasource locally using docker compose. Remember to set the `X-Scope-OrgID` header to `tenant1` in the Loki datasource in Perses.

# Screenshots
Screenshot 2026-02-19 at 21 36 11
Screenshot 2026-02-19 at 21 34 11
Screenshot 2026-02-19 at 21 36 17

# Checklist

## UI Changes


rickardsjp:
The go lint check seems erroneous:

```
level=error msg="Running error: context loading failed: no go files to analyze: running `go mod tidy` may solve the problem"
time="2026-02-19T22:09:47Z" level=error msg="issue with golangci for the plugin logexplorer" error="exit status 5"
```
If a purely frontend plugin like this one has no Go SDK, it shouldn't fail. If the maintainers agree, I can also open a PR to skip the linting if no `go.mod`/`go.sum` files are present?

rickardsjp:
> OK, I got the approval for an exclusion script. I will let you know when it is merged, then you can rebase main and the issue will disappear

What about https://github.com/perses/plugins/pull/578? Then we don't need to maintain a list.

Nexucis:
so, re-reading the issue raised and the proposition from each party I would say there is two different issues here @shahrokni.

The first one is: with the inclusion of the e2e tests package in the npm workspaces, we have now packages which are not plugins and that needs to be excluded from various scripts that are only running for the plugins.

Then you also have the issue we have here when we are checking golang code per plugins. This was fine because we never had a plugins with frontend only features (so here explorer). In the past, the explorer was included with the datasource, so we never go the issue until now.

On that matter, I think the fix proposed by @rickardsjp makes sense as indeed we don't want to support an exact list on what should be excluded for the go check.

But @rickardsjp we still want to have this check if there is a schema. So instead of verifying if you have go files, I think you should more check if you have the schema folder. If you have schema folder, then we want to have the go-sdk folder. So if the go files are not present, we want the script to fail.

Does it make sense for you @rickardsjp ?

clickhouse/src/queries/click-house-log-query/ClickHouseLogQuery.tsx
```diff
@@ -28,4 +29,18 @@ export const ClickHouseLogQuery: LogQueryPlugin<ClickHouseLogQuerySpec> = {
       variables: allVariables,
     };
   },
+  createVolumeQuery: (spec: ClickHouseLogQuerySpec): QueryDefinition | null => {
```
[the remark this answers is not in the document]
rickardsjp: Opened the PR on the shared repo: https://github.com/perses/shared/pull/103
And will adapt the log query plugins to use the shared interface.

Nexucis:
@rickardsjp any chance you can finish the work here. Having a generic explorer for the logs will be super interesting !

logexplorer/src/explore/log-explorer/LogExplorer.tsx
```diff
@@ -0,0 +1,205 @@
...
+            kind: 'Panel',
+            spec: { queries: queries, display: { name: '' }, plugin: { kind: 'LogsTable', spec: {} } },
+          }}
+        />
+      </DataQueriesProvider>
+    </Box>
+  );
+}
+
+function VolumeHistogramPanel({ queries }: { queries: QueryDefinition[] }): ReactElement {
```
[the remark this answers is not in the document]
rickardsjp: I think I would prefer to leave this as future work.

logexplorer/src/explore/log-explorer/LogExplorer.tsx
```diff
@@ -0,0 +1,205 @@
...
+        })
+        .map((p) => p.kind) ?? [],
+    [datasourcePlugins]
+  );
+
+  // State for volume queries
+  const [volumeQueries, setVolumeQueries] = useState<QueryDefinition[]>([]);
+
+  // Generate volume queries from log queries
+  useEffect(() => {
```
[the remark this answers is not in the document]
rickardsjp: `getPlugin` is async, so I think `useMemo` doesn't quite fit here and would return unresolved promises, no?

rickardsjp:
@Gladorme would you mind taking another look, please?

clickhouse/src/queries/click-house-log-query/click-house-log-query-types.test.ts
```diff
@@ -51,6 +51,7 @@ const createStubContext = (): ClickHouseQueryContext => {
       start: new Date('01-02-2025'),
     },
     variableState: {},
+    refreshKey: '',
```
Gladorme, line 54: I need to do a small cleanup, refreshKey is not needed anymore

logexplorer/src/explore/log-explorer/LogExplorer.tsx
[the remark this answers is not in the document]
Gladorme: You are right, however `useQuery` should do the work 🤔

logexplorer/.cjs.swcrc
```diff
@@ -0,0 +1,20 @@
+{
```
Gladorme: File not needed or I missed something?

logexplorer/.swcrc
```diff
@@ -0,0 +1,21 @@
+{
```
Gladorme: File not needed or I missed something?

logexplorer/.gitignore
```diff
@@ -0,0 +1,21 @@
+.idea/
```
Gladorme: File not needed or I missed something?

logexplorer/package.json
```diff
@@ -0,0 +1,71 @@
...
+    "url": "git+https://github.com/perses/plugins.git"
+  },
+  "bugs": {
+    "url": "https://github.com/perses/plugins/issues"
+  },
+  "scripts": {
+    "dev": "rsbuild dev",
+    "build": "npm run build-mf && concurrently \"npm:build:*\"",
+    "build-mf": "rsbuild build",
+    "build:cjs": "swc ./src -d dist/lib/cjs --strip-leading-paths --config-file .cjs.swcrc",
```
Gladorme:
```suggestion
    "build:cjs": "swc ./src -d dist/lib/cjs --strip-leading-paths --config-file .cjs.swcrc",
    "build:cjs": "swc ./src -d dist/lib/cjs --strip-leading-paths --config-file ../.cjs.swcrc",
    "build:esm": "swc ./src -d dist/lib --strip-leading-paths --config-file ../.swcrc",
```

Gladorme (commented):
Small comments about some files, the `useQuery` comment can be skipped. I prefer to not have useEffect in the code, but it look OK 😄 Else lgtm, tested manually with loki

logexplorer/.cjs.swcrc
rickardsjp: Yup, came with the original run of `percli plugin generate`. Removed, same for the ones below.

rickardsjp:
@Gladorme I would consider this ready for another round of review, PTAL when you get the chance.

Gladorme (approved):
LGTM, thanks for the PR!

PR #585, opened 2026-02-26
[ENHANCEMENT] TimeSeriesChart: add query name support for query settings

Gladorme:
# Description

Depends on: https://github.com/perses/shared/pull/69
Relates to https://github.com/perses/perses/issues/2469.

Add support for query name in query settings + deprecating queryIndex.

Migration guide:

- Via the UI:
Go the dashboard, edit timeserieschart panel, go to query settings, update any value and save (it will migrate all query settings of the panel to queryName)

- Manually:
Replace `queryIndex` by `queryName`.
Replace number by string `Query #<index + 1>` or use helper `defaultQueryName`.
Example: `queryIndex: 0` become `queryName: 'Query #1'`.

# Screenshots

Queries:
image

Query settings:
image

# Checklist

## UI Changes

### DIFF
--- a/timeserieschart/schemas/time-series.cue
+++ b/timeserieschart/schemas/time-series.cue
@@ -63,11 +63,11 @@ spec: close({
 
 #querySettings: [...{
 	queryName:    strings.MinRunes(1)
-	colorMode?:    "fixed" | "fixed-single"       // NB: "palette" could be added later
-	colorValue?:   =~"^#(?:[0-9a-fA-F]{3}){1,2}$" // hexadecimal color code
-	lineStyle?:    #lineStyle
-	areaOpacity?:  #areaOpacity
-	format?:       common.#format
+	colorMode?:   "fixed" | "fixed-single"       // NB: "palette" could be added later
+	colorValue?:  =~"^#(?:[0-9a-fA-F]{3}){1,2}$" // hexadecimal color code
+	lineStyle?:   #lineStyle
+	areaOpacity?: #areaOpacity
+	format?:      common.#format
 }]
 
 #lineStyle: "solid" | "dashed" | "dotted"
--- a/timeserieschart/schemas/time-series.cue
+++ b/timeserieschart/schemas/time-series.cue
@@ -14,6 +14,8 @@
 package model
 
 import (
+	"strings"
+
 	"github.com/perses/shared/cue/common"
 )
 
@@ -60,12 +62,12 @@ spec: close({
 }
 
 #querySettings: [...{
-	queryIndex:   int & >=0
-	colorMode?:   "fixed" | "fixed-single"       // NB: "palette" could be added later
-	colorValue?:  =~"^#(?:[0-9a-fA-F]{3}){1,2}$" // hexadecimal color code
-	lineStyle?:   #lineStyle
-	areaOpacity?: #areaOpacity
-	format?:      common.#format
+	queryName:    strings.MinRunes(1)
+	colorMode?:    "fixed" | "fixed-single"       // NB: "palette" could be added later
+	colorValue?:   =~"^#(?:[0-9a-fA-F]{3}){1,2}$" // hexadecimal color code
+	lineStyle?:    #lineStyle
+	areaOpacity?:  #areaOpacity
+	format?:       common.#format
 }]
 
 #lineStyle: "solid" | "dashed" | "dotted"
--- a/timeserieschart/schemas/time-series.cue
+++ b/timeserieschart/schemas/time-series.cue
@@ -62,7 +62,7 @@ spec: close({
 }
 
 #querySettings: [...{
-	queryName:    strings.MinRunes(1)
+	queryIndex:   strings.MinRunes(1)
 	colorMode?:   "fixed" | "fixed-single"       // NB: "palette" could be added later
 	colorValue?:  =~"^#(?:[0-9a-fA-F]{3}){1,2}$" // hexadecimal color code
 	lineStyle?:   #lineStyle
--- a/timeserieschart/schemas/time-series.cue
+++ b/timeserieschart/schemas/time-series.cue
@@ -62,7 +62,9 @@ spec: close({
 }
 
 #querySettings: [...{
-	queryIndex:   strings.MinRunes(1)
+	queryName?: strings.MinRunes(1)
+	// queryIndex is deprecated, use queryName instead. Kept for backward compatibility.
+	queryIndex?:  int & >=0
 	colorMode?:   "fixed" | "fixed-single"       // NB: "palette" could be added later
 	colorValue?:  =~"^#(?:[0-9a-fA-F]{3}){1,2}$" // hexadecimal color code
 	lineStyle?:   #lineStyle
--- a/timeserieschart/schemas/time-series.cue
+++ b/timeserieschart/schemas/time-series.cue
@@ -14,8 +14,6 @@
 package model
 
 import (
-	"strings"
-
 	"github.com/perses/shared/cue/common"
 )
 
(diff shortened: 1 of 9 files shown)


timeserieschart/src/utils/palette-gen.test.ts
```diff
@@ -165,7 +166,7 @@ describe('getSeriesColor', () => {
       seriesName: testSeriesName,
       seriesIndex: 0,
       querySettings: {
-        queryIndex: 0,
+        queryName: defaultQueryName(0),
```
Nexucis: that's a massive breaking change. This will break every dashboard created without dashboard as code or by migrating from Grafana.

Gladorme: Yeah, I will check if I can keep old data model

Gladorme: Ok, I remember why I did this change. We can't keep index, because it will cause issue when re-ordering queries with query settings. I will rename queryName to queryIndex, but result will be the same => they will need to update queryIndex.

timeserieschart/src/QuerySettingsEditor.tsx
```diff
@@ -42,35 +42,68 @@ import {
...
+}
+
+/**
+ * Normalizes a settings entry so it always persists via `queryName`.
+ * Converts the deprecated `queryIndex` to `queryName` and removes the old field.
+ */
+function normalizeToQueryName(querySettings: QuerySettingsOptions, queryNames: string[]): QuerySettingsOptions {
+  if (querySettings.queryIndex === undefined) {
+    return querySettings;
+  }
```
shahrokni: It seems that the idea behind this condition was that, if  `queryIndex`  was undefined then `queryName` would already exist. So, simply return the current one. However, this could go wrong. The `QuerySettingsOptions` object could be completely empty due to the fact that all fields are optional. Therefore, regardless of how queryName is generated, it is still technically possible to receive an empty object. This may open the door to bugs in future. 

```typeScript
/* ALL FIELDS ARE OPTIONAL */
export interface QuerySettingsOptions {
  queryName?: string;
  /**
   * @deprecated Use `queryName` instead.
   */
  queryIndex?: number;
  colorMode?: 'fixed' | 'fixed-single';
  colorValue?: string;
  lineStyle?: LineStyleType;
  areaOpacity?: number;
  format?: FormatOptions;
}
```

So, maybe?

```typescript
if (querySettings.queryIndex === undefined && querySettings.queryName!==undefined)
```

timeserieschart/src/time-series-chart-model.ts
```diff
@@ -40,7 +40,11 @@ export interface TimeSeriesChartOptions {
...
 export interface QuerySettingsOptions {
-  queryIndex: number;
+  queryName?: string;
+  /**
+   * @deprecated Use `queryName` instead.
+   */
+  queryIndex?: number;
   colorMode?: 'fixed' | 'fixed-single';
   colorValue?: string;
   lineStyle?: LineStyleType;
```
shahrokni, line 46: I understand that from the backend perspective a `none-optional`  `queryName` could be problematic and probably a breaking change.
 
```go
queryName?: strings.MinRunes(1)
```
However, from the UI perspective this field looks mandatory and there is no way to drop it. This means that from UI we could still keep it as a none-optional field.

Image

The fact that we could have empty object query settings is not a good idea in my opinion, although from the UI a `query name` is always guaranteed.  

```typescript
export interface QuerySettingsOptions {
  queryName?: string;
  /**
   * @deprecated Use `queryName` instead.
   */
  queryIndex?: number;
  colorMode?: 'fixed' | 'fixed-single';
  colorValue?: string;
  lineStyle?: LineStyleType;
  areaOpacity?: number;
  format?: FormatOptions;
}
```

timeserieschart/schemas/time-series.cue
```diff
@@ -60,7 +62,9 @@ spec: close({
 }
 
 #querySettings: [...{
-	queryIndex:   int & >=0
+	queryName?: strings.MinRunes(1)
+	// queryIndex is deprecated, use queryName instead. Kept for backward compatibility.
+	queryIndex?:  int & >=0
 	colorMode?:   "fixed" | "fixed-single"       // NB: "palette" could be added later
 	colorValue?:  =~"^#(?:[0-9a-fA-F]{3}){1,2}$" // hexadecimal color code
 	lineStyle?:   #lineStyle
```
shahrokni: Look at the following payload that I crafted and sent to the backend. 
Since the schema is fully optional now, such a payload is persisted. I think this should be avoided. If we insist that we should keep all fields optional, then maybe we should intercept the `req` and drop the `querySettings` completely. (I am just discussing the payload point of view)

Image

timeserieschart/src/time-series-chart-model.ts
Gladorme: Unfortunately, we can't enforce validation on cue on one field if one is missing and we don't want to do breaking change

timeserieschart/schemas/time-series.cue
Gladorme: Users could do it yes, but UI form enforce query name. So ok to me with limitation we have

timeserieschart/src/QuerySettingsEditor.tsx
Gladorme: Unfortunately, we can't enforce validation on cue on one field if one is missing and we don't want to do breaking change

timeserieschart/schemas/time-series.cue
shahrokni: > Users could do it yes, but UI form enforce query name. So ok to me with limitation we have

@jgbernalp Could we also have your opinion about this?

timeserieschart/src/utils/palette-gen.test.ts
jgbernalp: not sure why we are dropping the queryIndex, the index could be 0 and the name could be whatever. If name is a new field for better UX, queryIndex can be kept as it is.

timeserieschart/schemas/time-series.cue
jgbernalp: Since the JSON editor is editing directly you can do all sorts of things, including adding empty fields. IMO this is not an issue on the UI but on the backend validation, for optional fields there might be weird things allowed such as this example, but we probably cannot catch all the cases. I'd say we can improve the backend validation and unmarshalling in a different PR, so the resulting structure excludes empty objects if the schema says so.

Gladorme: With Cue, we can't enforce validation on at least one field :/

Gladorme: And it has not impact on the UI if empty object

timeserieschart/src/utils/palette-gen.test.ts
Gladorme: @jgbernalp Yes, queryIndex is not more dropped, but set as optional

jgbernalp: I wonder why it should be optional. Using the name to index might not be a good strategy, IIUC the index defines the order (functional) and the name is for display (cosmetic).

Gladorme: Now that query can have name, I thought it was a nice idea to be based on it. But the more I think about it, the more I think it's better to stick to index for now and keep name only for cosmetic purpose. 

The only drawback for query settings is: if you remove a query, it will change querysettings selector display. But it's already the case today, so I guess it's okay.

I will rollback to use queryIndex for functional purpose

Gladorme:
Should be good now @jgbernalp / @shahrokni. 

Btw, I think there a bug with query settings (already existing), if there are more than 1 series retuned by the query, the color is not applied (other styles are applied) or it's working as intended? 🤔

shahrokni:
> Should be good now @jgbernalp / @shahrokni.
> 
> Btw, I think there a bug with query settings (already existing), if there are more than 1 series retuned by the query, the color is not applied (other styles are applied) or it's working as intended? 🤔

I will take a look.

shahrokni (approved):

PR #632, opened 2026-04-16
[BUGFIX] PrometheusTimeSeriesQuery: Grafana migration: don't match non-prom queries when `type` is set

### LINKED ISSUE #4013: PrometheusTimeSeriesQuery: migration greedily matches non-Prometheus targets
## Problem

The [Prometheus query migration script](https://github.com/perses/plugins/blob/main/prometheus/schemas/prometheus-time-series-query/migrate/migrate.cue) matches any Grafana target containing an `expr` field, regardless of the target's datasource type. This is noted in the script itself:

> Best-effort conversion logic that may wrongly convert not-prometheus queries to PrometheusTimeSeriesQuery

This becomes a concrete problem when adding query migration scripts for other datasource plugins that also use `expr` (e.g. Loki). Since `percli` iterates query migration scripts via a Go map, iteration order is non-deterministic. If both a Prometheus and a Loki script match the same target, the result is a coin flip: sometimes `PrometheusTimeSeriesQuery`, sometimes `LokiLogQuery`.

## Investigation

We looked at two possible approaches to make the dispatch reliable:

### 1. Check `datasource.type` when available

Adding `type: "prometheus"` to the existing datasource constraint in the Prometheus script:

```cue
datasource?: {
    type: "prometheus"
    uid?: string
    ...
}
```

This works when `datasource.type` is present in the Grafana export. CUE rejects `type: "loki"` against the `type: "prometheus"` constraint, so Loki targets no longer match.

However, when `datasource.type` is absent (older dashboards, hand-crafted JSON), both scripts still match because `datasource?:` is optional and CUE fills in the constraint value on unification.

### 2. Analyze `expr` content to distinguish PromQL from LogQL

LogQL queries that use pipeline stages (`|=`, `|~`, `| json`, `| logfmt`, etc.) can be identified by a pipe operator after the stream selector closing brace:

```cue
expr: =~"^\\{[^}]*\\}\\s*\\|"
```

This correctly identifies LogQL with pipelines and rejects PromQL (including `{__name__="up"}`). However, bare LogQL stream selectors without pipes (e.g. `{container="nginx"}`) are syntactically identical to PromQL label matchers (e.g. `{__name__="up"}`), so they remain ambiguous.

### Conclusion

No single heuristic fully solves this. A layered approach covers the vast majority of real-world cases:

1. `datasource.type` present: Definitive match (handles most Grafana exports)
2. Pipe operator after `}`: Definitely LogQL
3. Starts with metric name or function: Definitely PromQL
4. Bare `{labels}` selector with no type info: Ambiguous between PromQL and LogQL; Prometheus fallback is acceptable since this is the current behavior

A combination of approach 1 (check `datasource.type` when available) and approach 2 (check for LogQL pipe syntax) would resolve the non-determinism for all practical cases.

Ref: perses/plugins#622

AntoineThebaud:
# Description

Closes https://github.com/perses/perses/issues/4013.

This PR reduces the "greediness" of the PrometheusTimeSeriesQuery migration logic by ensuring it's not matched when the datasource's `type` is explicitly set with another value than "prometheus". It doesn't change the current behavior when `type` is not set.

# Checklist

### DIFF
--- a/prometheus/schemas/prometheus-time-series-query/migrate/migrate.cue
+++ b/prometheus/schemas/prometheus-time-series-query/migrate/migrate.cue
@@ -14,13 +14,14 @@
 package migrate
 
 #target: {
-	// /!\ Best-effort conversion logic that may wrongly convert not-prometheus queries to PrometheusTimeSeriesQuery:
-	// Ideally we should rely on datasource.type = "prometheus" to identify prometheus queries. But in some cases,
-	// this information is not be available. Thus the condition relies on the presence of the "expr" field, that
-	// likely indicates that this is a prometheus query.
 	datasource?: {
-		uid: string
+		type?: "prometheus"
+		uid:   string
 	}
+	// /!\ Best-effort conversion logic that may wrongly convert not-prometheus queries to PrometheusTimeSeriesQuery:
+	// Ideally we should only rely on datasource.type = "prometheus" to identify prometheus queries. But in some cases,
+	// this information is not present. Thus, in addition to the check on the optional type above, the below condition
+	// relies on the presence of the "expr" field, that likely indicates that this is a prometheus query.
 	expr:          string
 	legendFormat?: string
 	interval?:     string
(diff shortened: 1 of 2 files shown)


ibakshay (approved):

rickardsjp:
Thanks @AntoineThebaud!

PR #634, opened 2026-04-20
[FEATURE] Loki: add query migration script for log queries

rickardsjp:
# Description

Add a Grafana migration script for Loki log queries (`loki-log-query`), enabling `percli migrate` to convert Grafana Loki targets to `LokiLogQuery`.

The script matches targets where `datasource.type` is `"loki"` and maps
- `expr` to `query`
- `datasource.uid` to datasource reference with `kind: LokiDatasource`

> [!NOTE]
> This PR covers panel targets with an explicit `datasource.type: "loki"`. Targets without type info may still be caught by the Prometheus migration script (https://github.com/perses/plugins/pull/632). The Prometheus migration still matches targets which have `expr` set but no explicit `type` set to Loki.

# Checklist

- [x] Pull request title follows the `[<catalog_entry>] <commit message>` naming convention.

## UI Changes

- n/a

### DIFF
--- a/loki/schemas/queries/loki-log-query/migrate/migrate.cue
+++ b/loki/schemas/queries/loki-log-query/migrate/migrate.cue
@@ -1,0 +1,34 @@
+// Copyright The Perses Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


loki/schemas/queries/loki-log-query/migrate/migrate.cue
```diff
@@ -0,0 +1,34 @@
...
+		type: "loki"
+		...
+	}
+	expr: string
+	...
+}
+
+kind: "LokiLogQuery"
+spec: {
+	if #target.datasource.uid != _|_ {
```
jgbernalp, line 27: since this is optional, we might be missing a test case when there is no `datasource.uid`

jgbernalp (approved):

PR #670, opened 2026-06-05
[ENHANCEMENT] timeserieschart: add per-query stack override

SoWieMarkus:
# Description

Adds a per-query `stack` override to the Time Series Chart plugin, allowing individual queries to opt in or out of stacking regardless of the global visual stack setting. This mirrors the behavior of Grafana/Plutono, where stacking can be controlled at the series level.

### What changed:
- Added a `stack?: bool` field to `QuerySettingsOptions` in the model, CUE schema, and Go SDK
- Added a `Stack` Switch in the `QuerySettingsEditor` 
- Updated `getTimeSeries()` in `data-transform.ts` to prefer the per-query `stack` flag over the global `visual.stack === 'all'` when set
- Moved `getCommonTimeScale` and `getTimeSeriesValues` imports from `@perses-dev/components` to `@perses-dev/core` (because not available in `@perses-dev/components`? Can revert that change ofc)

### Open question:
The drawing order of stacked series depends on the order queries are defined. This could be relevant now since time series can overlap. But if they are not stacked it's the same. So just a thought. A dedicated `zIndex` field in `QuerySettingsOptions` could give explicit control over this - though it only matters at higher opacity values anyway.

Also in Grafana / Plutono the Stacking Option is a boolean. I didn't want to adjust that since that would be a breaking change, but might also be something to consider.

# Screenshots

Query 1 excluded from stacking

Query 1 excluded from stacking.

Query 1 included again in stacking

Query 1 included again in stacking.

# Checklist

## UI Changes

### DIFF
--- a/timeserieschart/src/TimeSeriesChartBase.tsx
+++ b/timeserieschart/src/TimeSeriesChartBase.tsx
@@ -37,6 +37,7 @@ import {
   TooltipComponent,
 } from 'echarts/components';
 import { CanvasRenderer } from 'echarts/renderers';
+import { getCommonTimeScale } from '@perses-dev/core';
 import {
   ChartInstance,
   ChartInstanceFocusOpts,
@@ -48,7 +49,6 @@ import {
   enableDataZoom,
   FormatOptions,
   getClosestTimestamp,
-  getCommonTimeScale,
   getFormattedAxis,
   getPointInGrid,
   OnEventsType,
--- a/timeserieschart/src/TimeSeriesChartBase.tsx
+++ b/timeserieschart/src/TimeSeriesChartBase.tsx
@@ -37,7 +37,7 @@ import {
   TooltipComponent,
 } from 'echarts/components';
 import { CanvasRenderer } from 'echarts/renderers';
-import { getCommonTimeScale } from '@perses-dev/core';
+import { getCommonTimeScale } from '@perses-dev/components';
 import {
   ChartInstance,
   ChartInstanceFocusOpts,
--- a/timeserieschart/src/TimeSeriesChartBase.tsx
+++ b/timeserieschart/src/TimeSeriesChartBase.tsx
@@ -37,7 +37,6 @@ import {
   TooltipComponent,
 } from 'echarts/components';
 import { CanvasRenderer } from 'echarts/renderers';
-import { getCommonTimeScale } from '@perses-dev/components';
 import {
   ChartInstance,
   ChartInstanceFocusOpts,
@@ -49,6 +48,7 @@ import {
   enableDataZoom,
   FormatOptions,
   getClosestTimestamp,
+  getCommonTimeScale,
   getFormattedAxis,
   getPointInGrid,
   OnEventsType,
--- a/timeserieschart/src/TimeSeriesChartPanel.tsx
+++ b/timeserieschart/src/TimeSeriesChartPanel.tsx
@@ -450,8 +450,11 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps): ReactElement 
     setTimeRange({ start: new Date(event.start), end: new Date(event.end) });
   };
 
-  // Used to opt in to ECharts trigger item which show subgroup data accurately
-  const isStackedBar = visual.display === 'bar' && visual.stack === 'all';
+  // Used to opt in to ECharts trigger item which show subgroup data accurately.
+  // Derived from the actual series mapping rather than `visual.stack` alone so that
+  // bar charts stacked only via per-query overrides also use the right tooltip mode.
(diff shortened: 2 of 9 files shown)


timeserieschart/src/utils/data-transform.ts
```diff
@@ -13,14 +13,14 @@
 
 import type { YAXisComponentOption } from 'echarts';
 import { LineSeriesOption, BarSeriesOption } from 'echarts/charts';
+import { getCommonTimeScale } from '@perses-dev/core';
```
ibakshay:
```suggestion
import { getCommonTimeScale } from '@perses-dev/components';
```

timeserieschart/src/TimeSeriesChartBase.tsx
```diff
@@ -37,6 +37,7 @@ import {
   TooltipComponent,
 } from 'echarts/components';
 import { CanvasRenderer } from 'echarts/renderers';
+import { getCommonTimeScale } from '@perses-dev/core';
```
ibakshay:
```suggestion
import { getCommonTimeScale } from '@perses-dev/components';
```

timeserieschart/src/TimeSeriesChartPanel.tsx
```diff
@@ -42,8 +42,8 @@ import {
   DEFAULT_LEGEND,
   StepOptions,
   formatValue,
-  getTimeSeriesValues,
 } from '@perses-dev/components';
+import { getTimeSeriesValues } from '@perses-dev/core';
```
ibakshay:
```suggestion
import { getTimeSeriesValues } from '@perses-dev/components';
```

ibakshay (changes requested):
Great job with the PR! I left some minor comments. 

I also have a question. 
How does your change will affect the existing "Stack Series - All | None" feature? 

Image

timeserieschart/src/utils/data-transform.ts
```diff
@@ -90,7 +92,7 @@ export function getTimeSeries(
       datasetIndex,
       name: formattedName,
       color: paletteColor,
-      stack: visual.stack === 'all' ? visual.stack : undefined,
+      stack: shouldStack ? 'all' : undefined,
       yAxisIndex: yAxisIndex,
```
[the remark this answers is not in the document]
ibakshay: @SoWieMarkus, I can also confirm that the tooltip is not working anymore when it is stacked after your changes. I tested manually.

timeserieschart/src/utils/data-transform.ts
SoWieMarkus: I wondered: What is the best way to fix this locally on my machine. I couldn't find any docs about that. But I found the PR that changed that recently

SoWieMarkus:
> I also have a question. How does your change will affect the existing "Stack Series - All | None" feature?

@ibakshay The short answer: it doesn't change it. The per-query `stack` is purely an overlay on top of the existing global setting:

| Global `visual.stack` | Per-query `stack` | Result |
| --- | --- | --- |
| `'none'` / unset | unset | not stacked (unchanged) |
| `'none'` / unset | `true` | stacked (override, all queries that have that are stacked) |
| `'none'` / unset | `false` | not stacked |
| `'all'` | unset | stacked (unchanged) |
| `'all'` | `true` | stacked |
| `'all'` | `false` | not stacked (override) |

(I added test cases for that)

So any query that doesn't set the per-query `stack` continues to behave exactly as before. In theory you could change the `stack` option to a boolean like it is done in Grafana or Plutono, but I didn't want to change that because it would be a breaking change.

In theory you could further extend this feature by adding a `stacking groups` and then assign a query to one of this groups, but that is definitely out of scope for this PR though 😛

timeserieschart/src/utils/data-transform.ts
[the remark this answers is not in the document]
SoWieMarkus: Can you explain what exactly is not working anymore? I know there are issues with the tooltip when stacking is enabled, but that an issue that was there before my change.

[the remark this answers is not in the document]
ibakshay: > Can you explain what exactly is not working anymore? I know there are issues with the tooltip when stacking is enabled, but that an issue that was there before my change.

Yes, you are right. This bug is already existing. I will create an issue for it.

ibakshay:
@SoWieMarkus,  Your PR looks solid now! 
Could you please fix the lint issues?

ibakshay (approved):
LGTM! I have also tested manually and it works great! 
cc: @shahrokni , @AntoineThebaud, @jgbernalp

jgbernalp (approved):

PR #675, opened 2026-06-11
[BUGFIX] fix variable interpolation in prometheus label value query

zhuje:
# Description

Current (Broken) Format:
 `GET /api/v1/label/job/values?start=1781076832&end=1781080432&namespace=%24%7Bnamespace%3Aqueryparam%7D`
 Where `%24%7Bnamespace%3Aqueryparam%7D` is URL-encoded ${namespace:queryparam} (the raw, uninterpolated variable)

Expected (Fixed) Format:
`GET /api/v1/label/job/values?start=1781076832&end=1781080432&namespace=a&namespace=b&namespace=c&namespace=d`
 Where the namespace parameter is repeated for each selected value in the multi-select variable, just like the working query_range requests.

# Screenshots

Screenshot 2026-06-11 at 4 58 37 PM

# Checklist

## UI Changes

- [N/A] Changes that impact the UI include screenshots and/or screencasts of the relevant changes.


prometheus/src/plugins/PrometheusLabelValuesVariable.tsx
```diff
@@ -54,13 +54,15 @@ export const PrometheusLabelValuesVariable: VariablePlugin<PrometheusLabelValues
...
+  dependsOn: (spec: PrometheusLabelValuesVariableOptions, ctx: GetVariableOptionsContext) => {
     const matcherVariables = spec.matchers?.map((m) => parseVariables(m)).flat() || [];
     const labelVariables = parseVariables(spec.labelName);
     const datasourceVariables =
       spec.datasource && isVariableDatasource(spec.datasource) ? parseVariables(spec.datasource) : [];
+    const allAvailableVariables = Object.keys(ctx.variables || {});
+
     return {
-      variables: [...matcherVariables, ...labelVariables, ...datasourceVariables],
+      variables: [...matcherVariables, ...labelVariables, ...datasourceVariables, ...allAvailableVariables],
```
jgbernalp: I think this will cause a dependLock if 2 variables depend on each other, is this necessary?

jgbernalp: Checking in detail this causes an infinite loop when there is a variable dependent on another variable.

zhuje: Updated to extract just the queryParam variable

prometheus/src/plugins/PrometheusLabelValuesVariable.tsx
```diff
@@ -18,11 +18,59 @@ import {
...
+  }
+}
+
+function getQueryParamVariablesFromCache(
+  datasourceSelector: DatasourceSelector,
+  datasourceStore: DatasourceStore
+): string[] {
+  try {
+    const extendedStore = datasourceStore as ExtendedDatasourceStore;
+    if (!extendedStore.getDatasourceSpecSync) return [];
```
jgbernalp: AFAIC this function does not exist. It was added manually from your types.

zhuje: I needed to make changes in perses/shared so we could access the dashboard spec to get the queryParam variable:
https://github.com/perses/shared/pull/170/changes

prometheus/src/plugins/PrometheusLabelValuesVariable.tsx
```diff
@@ -18,11 +18,59 @@ import {
...
-import { PrometheusLabelValuesVariableOptions } from './types';
+import { PrometheusLabelValuesVariableOptions, PrometheusDatasourceSpec } from './types';
+
+interface ExtendedDatasourceStore extends DatasourceStore {
+  getDatasourceSpecSync?: (selector: DatasourceSelector) => DatasourceSpec | undefined;
+}
+
+function extractQueryParamVariables(datasourceSpec: DatasourceSpec<PrometheusDatasourceSpec>): string[] {
+  try {
+    const queryParams = datasourceSpec.plugin.spec.queryParams;
```
jgbernalp: This means we are excluding the interpolation for headers.

prometheus/src/plugins/PrometheusLabelValuesVariable.tsx
```diff
@@ -18,11 +18,59 @@ import {
...
+
+    if (!queryParams) return [];
+
+    const variables: string[] = [];
+    Object.values(queryParams).forEach((value) => {
+      if (typeof value === 'string') {
+        const variablesMap = parseVariablesAndFormat(value);
+        variablesMap.forEach((format, varName) => {
+          if (format === InterpolationFormat.QUERYPARAM) {
```
jgbernalp: other formats seem to be excluded

prometheus/src/plugins/PrometheusLabelValuesVariable.tsx
jgbernalp: ok next time we should reference the other PR in this PR description so the reviewer knows where the change comes from. This means we need a release in the shared libraries before we can merge and release this plugin.

jgbernalp (changes requested):

prometheus/src/plugins/PrometheusLabelValuesVariable.tsx
```diff
@@ -18,11 +18,94 @@ import {
...
+import { parseVariablesAndFormat } from '@perses-dev/components';
 import { DEFAULT_PROM, getPrometheusTimeRange, PROM_DATASOURCE_KIND } from '../model';
 import { stringArrayToVariableOptions, PrometheusLabelValuesVariableEditor } from './prometheus-variables';
 import { resolvePrometheusDatasource } from './interpolation';
-import { PrometheusLabelValuesVariableOptions } from './types';
+import { PrometheusLabelValuesVariableOptions, PrometheusDatasourceSpec } from './types';
+
+interface ExtendedDatasourceStore extends DatasourceStore {
+  getDatasourceSpecSync?: (selector: DatasourceSelector) => DatasourceSpec | undefined;
+}
```
jgbernalp: We should remove this as this will be present when we upgrade the shared dependency

prometheus/src/plugins/PrometheusLabelValuesVariable.tsx
```diff
@@ -18,11 +18,94 @@ import {
...
+    const variables: string[] = [];
+    const spec = datasourceSpec.plugin.spec;
+
+    // Helper function to extract variables from a string value
+    const extractFromString = (value: string): void => {
+      const variablesMap = parseVariablesAndFormat(value);
+      variablesMap.forEach((format, varName) => {
+        variables.push(varName);
+      });
+    };
```
jgbernalp: Is not a good practice to create functions that change global variables. Is better to create pure functions, in this case a function that returns the updated list.

```typescript
   const extractFromString = (value: string): string[] => {
      const variablesMap = parseVariablesAndFormat(value);
      return Array.from(variablesMap.keys())
    };
```

prometheus/src/plugins/PrometheusLabelValuesVariable.tsx
```diff
@@ -18,11 +18,94 @@ import {
...
-import { PrometheusLabelValuesVariableOptions } from './types';
+import { PrometheusLabelValuesVariableOptions, PrometheusDatasourceSpec } from './types';
+
+interface ExtendedDatasourceStore extends DatasourceStore {
+  getDatasourceSpecSync?: (selector: DatasourceSelector) => DatasourceSpec | undefined;
+}
+
+function extractDatasourceVariables(datasourceSpec: DatasourceSpec<PrometheusDatasourceSpec>): string[] {
+  try {
+    const variables: string[] = [];
```
jgbernalp:
```suggestion
    let variables: string[] = [];
```

prometheus/src/plugins/PrometheusLabelValuesVariable.tsx
```diff
@@ -18,11 +18,94 @@ import {
...
+    // Helper function to extract variables from a string value
+    const extractFromString = (value: string): void => {
+      const variablesMap = parseVariablesAndFormat(value);
+      variablesMap.forEach((format, varName) => {
+        variables.push(varName);
+      });
+    };
+
+    // Helper function to extract variables from string or array values
+    const extractFromValue = (value: string | string[]): void => {
```
jgbernalp:
```suggestion
    const extractFromValue = (value: string | string[]): string[] => {
```

prometheus/src/plugins/PrometheusLabelValuesVariable.tsx
zhuje: Updated in the last push to include interpolation for headers

prometheus/src/plugins/PrometheusLabelValuesVariable.tsx
zhuje: Update in the latest push to include other formats

prometheus/src/plugins/PrometheusLabelValuesVariable.tsx
zhuje: The CI might complain during the type-check, so we'll need to override the test.

prometheus/src/plugins/PrometheusLabelValuesVariable.tsx
```diff
@@ -18,11 +18,94 @@ import {
...
+}
+
+function extractDatasourceVariables(datasourceSpec: DatasourceSpec<PrometheusDatasourceSpec>): string[] {
+  try {
+    const variables: string[] = [];
+    const spec = datasourceSpec.plugin.spec;
+
+    // Helper function to extract variables from a string value
+    const extractFromString = (value: string): void => {
+      const variablesMap = parseVariablesAndFormat(value);
```
jgbernalp: can we use `parseVariables` instead?

zhuje: Updated to use parseVariables instead

prometheus/src/plugins/PrometheusLabelValuesVariable.tsx
zhuje: Updated the `string[]`

prometheus/src/plugins/PrometheusLabelValuesVariable.tsx
jgbernalp: Is fine that the test fails, it exactly fulfilling its purpose. We should not patch the test or create non existent types to make the CI happy. We need to address the shared library change, release and update.

prometheus/src/plugins/PrometheusLabelValuesVariable.tsx
zhuje: Updated to 
```
const allVariables: string[] = [];
... 
allVariables.push(...extractFromValue(value));
```

prometheus/src/plugins/PrometheusLabelValuesVariable.tsx
```diff
@@ -18,11 +18,87 @@ import {
...
-import { PrometheusLabelValuesVariableOptions } from './types';
+import { PrometheusLabelValuesVariableOptions, PrometheusDatasourceSpec } from './types';
+
+function extractDatasourceVariables(datasourceSpec: DatasourceSpec<PrometheusDatasourceSpec>): string[] {
+  try {
+    const spec = datasourceSpec.plugin.spec;
+
+    // Pure function to extract variables from a string value
+    const extractFromString = (value: string): string[] => {
+      return parseVariables(value);
```
jgbernalp: Do we need this function that is just wrapping another one?

prometheus/src/plugins/PrometheusLabelValuesVariable.tsx
```diff
@@ -18,11 +18,87 @@ import {
...
+    // Extract variables from queryParams
+    if (spec.queryParams) {
+      Object.values(spec.queryParams).forEach((value) => {
+        allVariables.push(...extractFromValue(value));
+      });
+    }
+
+    // Extract variables from directUrl
+    if (spec.directUrl) {
+      allVariables.push(...extractFromString(spec.directUrl));
```
jgbernalp:
```suggestion
      allVariables.push(...extractFromValue(spec.directUrl));
```

prometheus/src/plugins/PrometheusLabelValuesVariable.tsx
```diff
@@ -18,11 +18,87 @@ import {
...
+    }
+
+    // Extract variables from proxy configuration
+    if (spec.proxy?.spec) {
+      // Extract from proxy URL
+      if (spec.proxy.spec.url) {
+        allVariables.push(...extractFromString(spec.proxy.spec.url));
```
jgbernalp:
```suggestion
        allVariables.push(...extractFromValue(spec.proxy.spec.url));
```

prometheus/src/plugins/PrometheusLabelValuesVariable.tsx
```diff
@@ -18,11 +18,87 @@ import {
...
+      }
+
+      // Extract from proxy headers
+      if (spec.proxy.spec.headers) {
+        Object.values(spec.proxy.spec.headers).forEach((value) => {
+          if (typeof value === 'string') {
+            allVariables.push(...extractFromString(value));
```
jgbernalp:
```suggestion
            allVariables.push(...extractFromValue(value));
```

prometheus/src/plugins/PrometheusLabelValuesVariable.tsx
zhuje: True -- removed.

prometheus/src/plugins/PrometheusLabelValuesVariable.tsx
zhuje: Updated

prometheus/src/plugins/PrometheusLabelValuesVariable.tsx
zhuje: Updated

prometheus/src/plugins/PrometheusLabelValuesVariable.tsx
zhuje: Updated

jgbernalp (approved):

PR #682, opened 2026-06-16
[BUGFIX] Loki query migration incorrectly matching non-Loki datasources

gfierkasap:
# Description

This PR is one part of the solution for https://github.com/perses/perses/issues/4134

When migrating a Grafana dashboard to Perses, queries belonging to a Prometheus datasource were incorrectly migrated as LokiLogQuery with LokiDatasource.

This happened because the Loki migration script used a #target constraint to match targets with `datasource.type: "loki"`.  However, due to how the migration engine works — it fills `#target` with the Grafana JSON via CUE's FillPath before unifying with the script — the type: "loki" constraint is not enforced as a guard. Instead, CUE merges it into the open struct, so the script matches any target that has an expr field and a `datasource.uid`, regardless of whether `datasource.type` is present or correct.

This particularly affected dashboards where datasource.type is absent from the Grafana JSON (only uid is set), which is valid Grafana syntax. In those cases the Loki script would match and win the migration — non-deterministically, since migrateQuery iterates over a Go map.


loki/schemas/queries/loki-log-query/migrate/migrate.cue
```diff
@@ -15,20 +15,21 @@ package migrate
 
 #target: {
 	datasource: {
-		type: "loki"
-		...
+		type?: string
 	}
```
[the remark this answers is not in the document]
gfierkasap: [83933e0](https://github.com/perses/plugins/pull/682/commits/83933e00d1068a3e3a068edcc53ab253cf7a0e5c)

loki/schemas/queries/loki-log-query/migrate/migrate.cue
```diff
@@ -15,20 +15,21 @@ package migrate
...
 }
 
-kind: "LokiLogQuery"
-spec: {
-	if #target.datasource.uid != _|_ {
-		datasource: {
-			kind: "LokiDatasource"
-			name: #target.datasource.uid
+if #target.datasource.type != _|_ if #target.datasource.type == "loki" {
+    kind: "LokiLogQuery"
```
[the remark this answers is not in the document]
gfierkasap: [83933e0](https://github.com/perses/plugins/pull/682/commits/83933e00d1068a3e3a068edcc53ab253cf7a0e5c)

jgbernalp:
@gfierkasap thnx for your contribution. Would you mind adding tests for this case? We have test for these cases in other plugins in the form of JSON files that validate input and output. e.g. https://github.com/perses/plugins/tree/main/prometheus/schemas/prometheus-time-series-query/migrate/tests/basic

loki/schemas/queries/loki-log-query/migrate/migrate.cue
```diff
@@ -15,20 +15,22 @@ package migrate
 
 #target: {
 	datasource: {
-		type: "loki"
+		type?: string
```
Nexucis:
```suggestion
        type?: "loki"
        uid:   string
...
```

That sounds more accurate

gfierkasap: [c811fde](https://github.com/perses/plugins/pull/682/commits/c811fdecbe4c26f04c6038f302010ada89607d58)

gfierkasap:
> @gfierkasap thnx for your contribution. Would you mind adding tests for this case? We have test for these cases in other plugins in the form of JSON files that validate input and output. e.g. https://github.com/perses/plugins/tree/main/prometheus/schemas/prometheus-time-series-query/migrate/tests/basic

@jgbernalp I tried adding a test case covering this scenario, but I find adding a negative case (like this one) a bit problematic, since I assume the expected.json should be empty. I can keep working on it, so far all I get is an error on the test case with the message `Failed to parse expected.json: kind cannot be empty`

Nexucis (approved):
thank you for the fix. That will mitigate the issue but won't solve the entire issue unfortunately.
But that's mainly we are unable to discriminate the type of the query thanks to the poor grafana datamodel

Nexucis:
@gfierkasap can you format the file so we can move forward and merge this PR please ?
Thank you by advance.

loki/schemas/queries/loki-log-query/migrate/migrate.cue
```diff
@@ -15,20 +15,23 @@ package migrate
 
 #target: {
 	datasource: {
-		type: "loki"
+		type?: "loki"
+		uid: string
```
Nexucis: and copilot is right. Either `uid` is mandatory and then the check on UID is not necessary, either it is optional and so you need to reflect it in the type.

gfierkasap: @Nexucis So sorry for the late reply, missed the notification about the review on this PR. Anyhow, as far as I was able to confirm, type is required while uid is optional, updated it in [3f879f4](https://github.com/perses/plugins/pull/682/commits/3f879f4c41825a68dafedaaa42ba5a14edc6ca2f)

Nexucis (approved):
perfect, thank you @gfierkasap !

PR #688, opened 2026-06-18
[BUGFIX] fix cellSettings not applied when table column filters are active

rajusem:
## Summary

Fixes `cellSettings` conditional formatting (e.g., showing "N/A" for null values) not being applied when column filters are active in the Table panel.

### Root cause

`cellConfigs` was computed by iterating the unfiltered `data` array and keyed by array index (`${index}_${columnKey}`). The `<Table>` component receives `filteredData` which has different indices after filtering. This caused cellSettings overrides (text, colors) to either not appear or apply to incorrect cells when a filter was active.

### Fix

Moved `columnFilters` state and `filteredData` memo before the `cellConfigs` memo, then changed `cellConfigs` to iterate `filteredData` instead of `data`. This ensures cellConfig indices always match the rows the Table renders.

All hooks remain unconditional (early returns are after all hook calls), so React Rules of Hooks are preserved. Sorting is unaffected — TanStack Table preserves `row.id` from the original data index through `getSortedRowModel`, so `cellConfigs[cell.id]` resolves correctly regardless of sort order.

### How to reproduce (before fix)

1. Create a Table panel with multiple queries where some queries return data for a namespace and others don't (e.g., a multi-metric quota table)
2. Add `cellSettings` with a `Misc`/`null` condition (`text: "N/A"`)
3. Enable filtering
4. Verify N/A appears correctly in the unfiltered view
5. Apply a column filter to show only a row with null values
6. **Bug:** N/A text disappears — cells show blank instead

### Tests added

- N/A renders correctly for null values in unfiltered multi-query table
- N/A persists after filtering to a row with missing data (regression test)
- Genuine zero values are NOT replaced with N/A (false positive prevention)

Uses ARIA role-based selectors per [UI guidelines](https://github.com/perses/perses/blob/main/ui/ui-guidelines.md).

### Screenshots
Without filter :
Screenshot 2026-06-18 at 3 12 58 PM
With filter :
Screenshot 2026-06-18 at 11 52 12 AM

### Checklist


table/src/components/TablePanel.tsx
```diff
@@ -565,6 +565,31 @@ export function TablePanel({ contentDimensions, spec, queryResults }: TableProps
     return columns;
   }, [keys, spec.columnSettings, spec.defaultColumnHidden, allVariables, gaugeRangeByColumn, spec.cellSettings]);
 
+  // Filtering state — declared before cellConfigs so filteredData is available for cell config evaluation
+  const [columnFilters, setColumnFilters] = useState<ColumnFiltersState>([]);
+
+  // filter data based on the current filters
+  const filteredData = useMemo(() => {
+    let filtered = [...data];
```
jgbernalp: This will create a new reference each time, making unnecessary re renders when there are no filters. We should return the data as it is if there are no filters, otherwise copy

table/src/components/TablePanel.test.tsx
```diff
@@ -351,4 +354,118 @@ describe('TablePanel', () => {
...
+        // ns-a should have both values — verify value #2 = 50 is NOT shown as N/A
+        expect(await screen.findByRole('cell', { name: 'ns-a' })).toBeInTheDocument();
+        expect(await screen.findByRole('cell', { name: '50' })).toBeInTheDocument();
+
+        // ns-b should show N/A for the missing value #2
+        expect(await screen.findByRole('cell', { name: 'ns-b' })).toBeInTheDocument();
+        const naCells = await screen.findAllByRole('cell', { name: 'N/A' });
+        expect(naCells.length).toBeGreaterThanOrEqual(1);
+
+        // Verify genuine zero (value #1 = 0) would NOT become N/A — 0 is a valid number, not null
```
jgbernalp: Is this comment being tested, it seems the assertions do not include the check for `0`

table/src/components/TablePanel.tsx
rajusem: Good catch! Fixed in 5b8b9df — `filteredData` now returns `data` directly when no filters are active, avoiding unnecessary reference changes that would cascade recomputation to `cellConfigs`.

table/src/components/TablePanel.test.tsx
rajusem: The zero-value assertion exists — it was just a misplaced comment. The check is in a separate dedicated test `should NOT show N/A for genuine zero values` which uses `MOCK_MULTI_QUERY_DATA_WITH_ZERO` and explicitly asserts that `0` renders as `0` and no N/A appears. Removed the misleading comment from this test in 5b8b9df.

jgbernalp (approved):

PR #712, opened 2026-06-29
[ENHANCEMENT] ScatterChart: make axis and tooltip timezone aware

andreasgerstmayr:
# Description

scatterchart: make axis and tooltip timezone aware

# Screenshots

No UI changes, except that the timezone from the TimeRangeSelector is respected.

# Checklist

## UI Changes

### DIFF
--- a/scatterchart/src/utils/timezone-formatter.ts
+++ b/scatterchart/src/utils/timezone-formatter.ts
@@ -1,0 +1,49 @@
+// Copyright The Perses Authors
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+// http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import { formatWithTimeZone } from '@perses-dev/components';
+
+const DAY_MS = 86400000;
+const MONTH_MS = 2629440000;
+const YEAR_MS = 31536000000;
+
+/**
+ * Creates a timezone-aware axis formatter function for different time ranges
+ */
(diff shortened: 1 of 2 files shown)


scatterchart/src/utils/timezone-formatter.ts
```diff
@@ -0,0 +1,49 @@
...
+    if (rangeMs > YEAR_MS * 5) {
+      return formatWithTimeZone(timeStamp, 'yyyy', timeZone);
+    }
+
+    // more than 2 years
+    if (rangeMs > YEAR_MS * 2) {
+      return formatWithTimeZone(timeStamp, 'MMM yyyy', timeZone);
+    }
+
+    // between 10 days to 6 months
```
celian-garcia: What happen between 2 years and 6 month ?

celian-garcia: Can you still make a screenshot for this ? if not to complicate

andreasgerstmayr: This is the identical axis formatter as used in the timeseries, heatmap and statushistory charts:
https://github.com/perses/plugins/blob/39a7700a8fd33c1539a1af12b828db35bcdd91d0/timeserieschart/src/utils/timezone-formatter.ts
https://github.com/perses/plugins/blob/39a7700a8fd33c1539a1af12b828db35bcdd91d0/heatmapchart/src/utils/get-formatted-axis-label.ts
https://github.com/perses/plugins/blob/39a7700a8fd33c1539a1af12b828db35bcdd91d0/statushistorychart/src/utils/get-formatted-axis-label.ts

It's a way to make the axis timezone-aware. If the range of the chart is too long it uses years, otherwise months, days, hours etc.

Probably a candidate to move to https://github.com/perses/shared/tree/main/components/src/EChart (including the bug fix below)?

Screenshot:


> What happen between 2 years and 6 month ?

I think you just discovered a bug in all these charts :D
You're right, it'll fall back to the return statement, which formats it as `HH:mm`. I fixed it now in this file.

andreasgerstmayr: I created https://github.com/perses/shared/pull/190 to pull this shared code in the `perses/shared` repo. But we'll need a new release before I can remove it from the plugins here.

celian-garcia (approved):
Lgtm. The update for alignment can be done later indeed

PR #725, opened 2026-07-09
[BUGFIX] pyroscope: fix timerange

Gladorme:
# Description

Fix pyroscope timerange query params: https://github.com/grafana/pyroscope/blob/254c6885170b4ac35d3c8933ffdbfd2f955fade2/api/types/v1/types.proto#L64

Improving the display on the way too. When changing datasource, input options were not updated, because client is "enabled", but on old spec so the query is executed but with wrong data. That why I put client as useQuery key instead of datasource selector. I would prefer to use datasource name as key, but it may need more work.

# Screenshots

# Checklist

## UI Changes

### DIFF
--- a/pyroscope/src/components/Service.tsx
+++ b/pyroscope/src/components/Service.tsx
@@ -24,7 +24,6 @@ export interface ServiceProps {
 
 export function Service(props: ServiceProps): ReactElement {
   const { datasource, value, onChange } = props;
-
   const { data: servicesOptions, isLoading: isServicesOptionsLoading } = useServices(datasource);
 
   function handleServiceChange(_event: SyntheticEvent, newValue: string | null): void {
--- a/pyroscope/src/components/Service.tsx
+++ b/pyroscope/src/components/Service.tsx
@@ -22,8 +22,7 @@ export interface ServiceProps {
   onChange?(value: string): void;
 }
 
-export function Service(props: ServiceProps): ReactElement {
-  const { datasource, value, onChange } = props;
+export function Service({ datasource, value, onChange }: ServiceProps): ReactElement {
   const { data: servicesOptions, isLoading: isServicesOptionsLoading } = useServices(datasource);
 
   function handleServiceChange(_event: SyntheticEvent, newValue: string | null): void {
--- a/pyroscope/src/model/profile-query-model.ts
+++ b/pyroscope/src/model/profile-query-model.ts
@@ -23,3 +23,13 @@ export interface PyroscopeProfileQuerySpec {
   filters?: LabelFilter[];
   service?: string;
 }
+
+/**
+ * A Pyroscope profile query needs at least a service and a profile type to be executable.
+ * When either is missing, the query is considered incomplete and should not be run.
+ */
+export function isProfileQueryComplete(
+  spec: Partial<Pick<PyroscopeProfileQuerySpec, 'service' | 'profileType'>>
+): boolean {
+  return !!spec.service && !!spec.profileType;
+}
--- a/pyroscope/src/utils/use-query.ts
+++ b/pyroscope/src/utils/use-query.ts
@@ -26,12 +26,18 @@ import { getUnixTimeRange } from '../plugins';
 
 export function useLabelNames(datasource: DatasourceSelector): UseQueryResult<SearchLabelNamesResponse, StatusError> {
   const { data: client } = useDatasourceClient<PyroscopeClient>(datasource);
+  const { absoluteTimeRange } = useTimeRange();
+  const { start, end } = getUnixTimeRange(absoluteTimeRange);
 
   return useQuery<SearchLabelNamesResponse, StatusError>({
     enabled: !!client,
     queryKey: ['searchLabelNames', 'datasource', datasource],
     queryFn: async () => {
-      return await client!.searchLabelNames({}, { 'content-type': 'application/json' }, {});
+      return await client!.searchLabelNames(
+        {},
+        { 'content-type': 'application/json' },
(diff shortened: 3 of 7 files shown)


prakhar29jain (approved):

pyroscope/src/components/Service.tsx
```diff
@@ -24,7 +24,6 @@ export interface ServiceProps {
 
 export function Service(props: ServiceProps): ReactElement {
   const { datasource, value, onChange } = props;
-
   const { data: servicesOptions, isLoading: isServicesOptionsLoading } = useServices(datasource);
```
shahrokni: Was removing the space between the two lines intentional?

pyroscope/src/model/profile-query-model.ts
```diff
@@ -23,3 +23,13 @@ export interface PyroscopeProfileQuerySpec {
...
+
+/**
+ * A Pyroscope profile query needs at least a service and a profile type to be executable.
+ * When either is missing, the query is considered incomplete and should not be run.
+ */
+export function isProfileQueryComplete(
+  spec: Partial<Pick<PyroscopeProfileQuerySpec, 'service' | 'profileType'>>
+): boolean {
+  return !!spec.service && !!spec.profileType;
+}
```
shahrokni, line 35: It seems `profileType` is a none-optional field of `PyroscopeProfileQuerySpec`, meaning it is expected to always have a value unless it is **string empty**.  Could you please explain, if it would be possible to save string empty into the profileType? (According to the schema we can, then the question would be why?)

Let's look at the schema,
Maybe we should avoid that in the schema ... 

```go
kind: "PyroscopeProfileQuery"
spec: close({
	ds.#selector
	maxNodes?:   number
	profileType: string
	filters?: [...{
		labelName:  string
		labelValue: string
		operator:   string
	}]
	service?: string
})
```

pyroscope/src/utils/use-query.ts
```diff
@@ -46,12 +52,12 @@ export function useLabelValues(
...
   return useQuery<SearchLabelValuesResponse, StatusError>({
     enabled: !!client,
-    queryKey: ['searchLabelValues', labelName, 'datasource', datasource],
+    queryKey: ['searchLabelValues', labelName, client],
     queryFn: async () => {
       return await client!.searchLabelValues(
         {},
         { 'content-type': 'application/json' },
-        { name: labelName, from: start, until: end }
+        { name: labelName, start: start * 1000, end: end * 1000 }
```
Nexucis: that will be worth to add a comment explaining why you are adding a `* 1000` everytime. It's not really explanatory to have this factor

pyroscope/src/utils/use-query.ts
```diff
@@ -26,12 +26,18 @@ import { getUnixTimeRange } from '../plugins';
...
     enabled: !!client,
-    queryKey: ['searchLabelNames', 'datasource', datasource],
+    queryKey: ['searchLabelNames', client],
     queryFn: async () => {
-      return await client!.searchLabelNames({}, { 'content-type': 'application/json' }, {});
+      return await client!.searchLabelNames(
+        {},
+        { 'content-type': 'application/json' },
+        { start: start * 1000, end: end * 1000 }
+      );
```
shahrokni, line 43: I did some research (not 100% sure it is valid though), that if you pass an object to `queryKey` it would serialize it and use it as a part of that key. 

**IF THIS ASSUMPTION IS CORRECT, THEN**

**_Now, lets see if using client would really solve this issue_**
 a data source selector is 

```typescript
export interface DatasourceSelector {
    /**
     * Kind of the datasource.
     */
    kind: string;
    /**
     * Name of the datasource.
     * If omitted, it's assumed that you target the default datasource for the specified kind (and group, if set)
     */
    name?: string;
}
```

 and Pyroscop client is 

```typescript
interface PyroscopeClientOptions {
  datasourceUrl: string;
  headers?: RequestHeaders;
}

export interface PyroscopeClient extends DatasourceClient {
  options: PyroscopeClientOptions;
  searchProfiles(params: SearchProfilesParameters, headers?: RequestHeaders): Promise<SearchProfilesResponse>;
  searchProfileTypes(
    params: SearchProfileTypesParameters,
    headers: RequestHeaders,
    body: Record<string, string | number>
  ): Promise<SearchProfileTypesResponse>;
  searchLabelNames(
    params: SearchLabelNamesParameters,
    headers: RequestHeaders,
    body: Record<string, string | number>
  ): Promise<SearchLabelNamesResponse>;
  searchLabelValues(
    params: SearchLabelValuesParameters,
    headers: RequestHeaders,
    body: Record<string, string | number>
  ): Promise<SearchLabelValuesResponse>;
  searchServices(
    params: SearchLabelValuesParameters,
    headers: RequestHeaders,
    body: Record<string, string | number>
  ): Promise<SearchLabelValuesResponse>;
}
```
What if I create two pyroscope data source with the exact same configuration, (keep the name empty)
Then would this key still work when switching between the two? 


Image

pyroscope/src/components/Service.tsx
Gladorme: Yes preferences, and I changed again to remove one line :)

pyroscope/src/model/profile-query-model.ts
Gladorme: Enforced string with `strings.MinRunes(1)`

pyroscope/src/utils/use-query.ts
Gladorme: Yes, but if they have same config, no need to update data, so good :)

shahrokni (approved):

PR #726, opened 2026-07-10
[BREAKINGCHANGE] Tempo: Use Query V2 API and surface API (warning) messages

### LINKED ISSUE #3961: Tempo: support the v2 of the API
Tempo plugin does not support yet the v2 API of search tag and traces which are latest in nature

https://grafana.com/docs/tempo/latest/api_docs/

andreasgerstmayr:
# BREAKING CHANGE
This PR changes the query endpoint to use `/api/v2/traces` instead of `/api/traces`.
**If you have existing Tempo datasources, please update the list of allowed endpoints: Replace `/api/traces` with `/api/v2/traces`**

# Description

Use Query V2 API and surface API (warning) messages.
Existing datasources may only have /api/traces in their allowed endpoints list, therefore fall back to the v1 API on 403 errors for backwards compatibility.

Resolves: https://github.com/perses/perses/issues/3961

# Screenshots

Example notice:
image

# Checklist

## UI Changes

### DIFF
--- a/tempo/src/model/tempo-client.ts
+++ b/tempo/src/model/tempo-client.ts
@@ -12,7 +12,7 @@
 // limitations under the License.
 
 import { DatasourceClient } from '@perses-dev/plugin-system';
-import { RequestHeaders } from '@perses-dev/client';
+import { fetchJson, RequestHeaders, UserFriendlyError } from '@perses-dev/client';
 import * as otlptracev1 from '@perses-dev/spec/dist/dashboard/query-type/otlp/trace/v1/trace';
 import {
   QueryRequestParameters,
@@ -46,17 +46,7 @@ export interface QueryOptions {
   headers?: RequestHeaders;
 }
 
-export const executeRequest = async <T>(...args: Parameters<typeof global.fetch>): Promise<T> => {
-  const response = await fetch(...args);
-  try {
-    return await response.json();
-  } catch (e) {
-    console.error('Invalid response from server', e);
-    throw new Error('Invalid response from server');
-  }
-};
-
-function fetchWithGet<TRequest extends RequestParams<TRequest>, TResponse>(
+async function fetchWithGet<TRequest extends RequestParams<TRequest>, TResponse>(
   apiURI: string,
   params: TRequest,
   queryOptions: QueryOptions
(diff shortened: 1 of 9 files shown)


tempo/src/model/tempo-client.ts
```diff
@@ -102,17 +104,46 @@ export function search(params: SearchRequestParameters, queryOptions: QueryOptio
...
+export async function query(params: QueryRequestParameters, queryOptions: QueryOptions): Promise<QueryResponse> {
+  try {
+    const response = await fetchWithGet<Record<string, never>, QueryResponse>(
+      `/api/v2/traces/${encodeURIComponent(params.traceId)}`,
+      {},
+      queryOptions
+    );
+
+    // Unlike /api/traces, /api/v2/traces returns an empty trace instead of returning a 404 error.
+    if (!response.trace.resourceSpans) {
```
jgbernalp:
```suggestion
    if (!response.trace?.resourceSpans) {
```

tempo/src/model/tempo-client.ts
```diff
@@ -102,17 +104,46 @@ export function search(params: SearchRequestParameters, queryOptions: QueryOptio
...
+      throw new UserFriendlyError('Trace not found', 404);
+    }
+    return response;
+  } catch (e) {
+    // Existing datasources may only have /api/traces in their allowed endpoints list.
+    // Fall back to the v1 API on 403 errors for backwards compatibility.
+    if (e instanceof Error && 'status' in e && e.status === 403) {
```
jgbernalp: This fallback triggers on any 403, so genuine auth failures will also be caught here. Would it be better to allow the client to be configured to use a specific version?

andreasgerstmayr: I would like to avoid the additional complexity of introducing a data source configuration setting for this setting. Potentially in the future there will be a V2 of the `search` endpoint, then we'll end up with two boolean configs "enable query v2" and "enable search v2".

For the UI path, I could enable the new API by default, and keep existing datasources as-is. But when the datasource is provisioned via the operator, I also would like to enable the new API by default, however I cannot differentiate between new datasources (where this config property is unset) and existing datasources (where this config property is also unset). Unless we create some upgrade/migration path (enable v1 API for existing datasources), but that again adds complexity.

Ideally I'd also remove the fallback-to-v1-on-403-error workaround after some time (a year?).

What do you think?

jgbernalp: I think mixing the 2 versions might lead to issues other than the one with the 403. I'd say is better to default to the new version, and if existing users still want v1 then we should allow something like `query: 'v1'` so they can use the v1 client, then we can mark this one as a Breaking change. WDYT @Nexucis ?

andreasgerstmayr: The Query V2 API was added in Tempo v2.6.0 (released Sep 3, 2024), I don't think there's a good reason to keep supporting the old `v1` API.

afaics the only issue is the allowed endpoints list, which will prevent connections to the new API for existing Tempo datasources. But if we mark it as a breaking change, I guess it should be ok (the fix is simple, to update the list of allowed endpoints in the datasource).

jgbernalp: In that case we can remove the catch code falling back to v1 and create a new minor version for the plugin, marking the breaking change and adding in the PR small documentation on how to "migrate".

andreasgerstmayr: I've removed the fallback, added migration steps to the description, and bumped the version.

jgbernalp (approved):

PR #736, opened 2026-07-19
[BUGFIX] Table: use last data point instead of first in buildRawTableData

tgitelman:
buildRawTableData() hardcodes ts.values[0] — always picking the first data point from range query results. When a datasource plugin sends query_range despite the table requesting instant mode, the first step covers a stale time window while the last step is current. StatChart panels (using calculation: last) show correct values; the table does not.

Fix: use ts.values[ts.values.length - 1] instead of ts.values[0] for value extraction, timestamp, and CSV export. Also add a guard for empty values arrays.

Ref: perses/perses#4270

# Description

# Screenshots

# Checklist

## UI Changes


jgbernalp (approved):

Nexucis:
type-check failed. Can you check @tgitelman please

Nexucis (approved):

PR #737, opened 2026-07-19
[BUGFIX] Loki: dispatch to instant query when context.mode is instant

tgitelman:
## Summary

Ref: perses/perses#4270

`getLokiTimeSeriesData()` always calls `client.queryRange()` regardless of `context.mode`. When a panel requests instant mode (e.g. Table without column plugins), the Loki plugin ignores it and sends `query_range` — unlike the Prometheus plugin which checks `context.mode` and dispatches to `instantQuery()` vs `rangeQuery()`.

This also causes empty results for large time ranges (14d+): `query_range` with large range vectors triggers Loki's `split_queries_by_interval`, which truncates lookback windows. The instant endpoint (`/loki/api/v1/query`) does not split and evaluates correctly.

### Changes

**`loki/src/queries/loki-time-series-query/get-loki-time-series-data.ts`:**
- Check `context.mode` — dispatch to `client.query()` (`/loki/api/v1/query`) for instant mode
- Added `convertVectorToTimeSeries()` to convert instant query vector responses to the `TimeSeries` format
- `streams` resultType (raw log queries without aggregation) returns empty — not applicable for time series panels
- Range query path completely unchanged

**`loki/src/queries/loki-time-series-query/loki-time-series-query-types.ts`:**
- Updated `LokiTimeSeriesQueryResponse` to include `LokiQueryResponse` (instant query responses)

### Reproduction

1. Create a dashboard with a LokiDatasource
2. Add a Table panel (no column plugins) with `count_over_time(... [$__range])`
3. Enable Loki gateway debug logs (`--log.level=debug`)
4. **Before fix:** 100% of requests are `/loki/api/v1/query_range`, zero instant queries
5. **After fix:** Table requests go to `/loki/api/v1/query`
6. Select "Last 14 days" — **before fix:** empty results; **after fix:** correct totals

### Test plan

- [ ] Loki gateway logs show `/loki/api/v1/query` (instant) requests when Table panel loads
- [ ] Large time ranges (14d+) return correct data instead of empty results
- [ ] Range queries (time series charts) continue working unchanged
- [ ] Aggregation queries (`count_over_time`, `sum_over_time`) return correct single values
- [ ] Raw log queries (no aggregation) in instant mode return empty gracefully

### DIFF
--- a/loki/src/queries/loki-time-series-query/get-loki-time-series-data.ts
+++ b/loki/src/queries/loki-time-series-query/get-loki-time-series-data.ts
@@ -14,8 +14,8 @@
 import { TimeSeriesQueryPlugin, replaceVariables } from '@perses-dev/plugin-system';
 import { milliseconds } from 'date-fns';
 import { DurationString, parseDurationString, TimeSeries } from '@perses-dev/spec';
-import { LokiClient } from '../../model/loki-client';
-import { LokiMatrixResult } from '../../model/loki-client-types';
+import { LokiClient, toUnixSeconds } from '../../model/loki-client';
+import { LokiMatrixResult, LokiVectorResult } from '../../model/loki-client-types';
 import { DEFAULT_DATASOURCE } from '../constants';
 import { LokiTimeSeriesQuerySpec } from './loki-time-series-query-types';
 
@@ -88,6 +88,19 @@ function convertMatrixToTimeSeries(matrix: LokiMatrixResult[]): TimeSeries[] {
   });
 }
 
+function convertVectorToTimeSeries(vector: LokiVectorResult[]): TimeSeries[] {
+  return vector.map((series) => {
+    const name = Object.entries(series.metric)
+      .map(([k, v]) => `${k}=${v}`)
+      .join(', ');
+    return {
+      name,
+      values: [[Number(series.value[0]) * 1000, Number(series.value[1])]],
+      labels: series.metric,
+    };
+  });
+}
+
 export const getLokiTimeSeriesData: TimeSeriesQueryPlugin<LokiTimeSeriesQuerySpec>['getTimeSeriesData'] = async (
   spec,
   context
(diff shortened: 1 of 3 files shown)


loki/src/queries/loki-time-series-query/get-loki-time-series-data.ts
```diff
@@ -107,7 +120,31 @@ export const getLokiTimeSeriesData: TimeSeriesQueryPlugin<LokiTimeSeriesQuerySpe
 
   const { start, end } = context.timeRange;
 
-  // Calculate proper step using similar logic to Prometheus
+  if (context.mode === 'instant') {
+    // Convert to Unix seconds — client.query() passes time as-is
+    // (unlike queryRange which normalizes internally via toUnixSeconds)
+    const time = Math.floor(end.getTime() / 1000).toString();
+    const response = await client.query({ query, time });
```
jgbernalp:
```suggestion
    const response = await client.query({ query, time: toUnixSeconds(end) });
```

loki/src/queries/loki-time-series-query/get-loki-time-series-data.ts
```diff
@@ -107,7 +120,31 @@ export const getLokiTimeSeriesData: TimeSeriesQueryPlugin<LokiTimeSeriesQuerySpe
 
   const { start, end } = context.timeRange;
 
-  // Calculate proper step using similar logic to Prometheus
+  if (context.mode === 'instant') {
```
jgbernalp, line 123: We should add some tests for this case

loki/src/queries/loki-time-series-query/get-loki-time-series-data.ts
```diff
@@ -107,7 +120,31 @@ export const getLokiTimeSeriesData: TimeSeriesQueryPlugin<LokiTimeSeriesQuerySpe
...
+        series: convertVectorToTimeSeries(response.data.result as LokiVectorResult[]),
+        timeRange: { start, end },
+        stepMs: DEFAULT_MIN_STEP_SECONDS * 1000,
+        metadata: { executedQueryString: query },
+      };
+    }
+
+    // 'streams' resultType: raw log queries without aggregation.
+    // Not applicable for time series panels — return empty.
+    return {
```
jgbernalp, line 135: We should add a notice, to inform the user why the data is empty in this case.
```suggestion
    return {
       metadata: { notices: [{ type: 'warning', message: "log streams are not supported in 'instant' mode"] },
```

loki/src/queries/loki-time-series-query/get-loki-time-series-data.ts
tgitelman: Applied

loki/src/queries/loki-time-series-query/get-loki-time-series-data.ts
tgitelman: Implemented.

loki/src/queries/loki-time-series-query/get-loki-time-series-data.ts
tgitelman: Applied

jgbernalp (approved):

PR #739, opened 2026-07-20
[IGNORE] fix vulnerable dependencies

jgbernalp:
# Description

Lock react version for development to 18 to avoid react 19 resolutions. Fix transitive vulnerable dependencies

# Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -87,6 +87,8 @@
     "react-router-dom": "^6.30.3",
     "rsbuild-plugin-dts": "^0.18.0",
     "ts-node": "^10.9.2",
+    "react": "^18.3.1",
+    "react-dom": "^18.3.1",
     "turbo": "^2.6.1",
     "typescript": "^5.4.2"
   },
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -87,6 +87,8 @@
     "react-router-dom": "^6.30.3",
     "rsbuild-plugin-dts": "^0.18.0",
     "ts-node": "^10.9.2",
+    "react": "^18.3.1",
```
Gladorme, line 90: What about React 17? It will not raise issue? 🤔

jgbernalp: no is a dev dependency. The peer still has react 17.

Gladorme (approved):

package.json
Gladorme: Looks strange, but I guess it's for forcing version upgrade?

jgbernalp: when running npm audit, it tries to update to react 19 as this is defined in material UI as peer dependency, when this happens it selects the latest version. Locking to react 18 allows to set the specific resolution so the audit does not conclict.

PR #747, opened 2026-07-22
[ENHANCEMENT] pyroscope: optimize transformProfileResponse to O(n)

luweglarz:
# Description

There was an explicit todo on this function to optimize from O(n2) to O(n).

The old code rescanned the whole parent level from scratch for every node. Since both levels are already sorted by `start`, a single pointer that only moves forward is enough in one pass.

# Checklist

### DIFF
--- a/pyroscope/src/plugins/pyroscope-profile-query/get-profile-data.ts
+++ b/pyroscope/src/plugins/pyroscope-profile-query/get-profile-data.ts
@@ -90,145 +90,102 @@ export const getProfileData: ProfileQueryPlugin<PyroscopeProfileQuerySpec>['getP
   return transformProfileResponse(response);
 };
 
-/**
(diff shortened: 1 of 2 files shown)


pyroscope/src/plugins/pyroscope-profile-query/get-profile-data.ts
jgbernalp, line 1: It would be good to add a test for this, so we can ensure it does not affect the output.

luweglarz: Is a comparison of outputs with the legacy function and the new one in test cases is what you mean?

luweglarz: I added a set of test cases for the refactor + regression test cases against old implementation

jgbernalp (approved):

PR #753, opened 2026-07-24
[BUGFIX] Timeserieschart: default query color mode to fixed for all series

### LINKED ISSUE #4204: TimeseriesChart: query settings color not applied if query return more than one series
### What did you do?

If you define a query setting with specific color and the associated query returns more than 1 series. The color is not applied, but other query settings are applied (dash, dot, ... display)

### What did you expect to see?

I expect to have all the series to be in the configured color as defined in the query settings if there is one.

### Perses version

0.54.0-beta.1

mixturesolutions:
# Description

Fixes [perses/perses#4204](https://github.com/perses/perses/issues/4204).
When adding a query color setting, the UI defaulted to `fixed-single`, which only applies the color if the query returns one series. Other query settings (line style, opacity, etc.) always apply to every series from that query, so multi-series queries looked inconsistent: dash/dot applied, but color did not.
This changes the default color mode to `fixed`, so the configured color applies to all series returned by the query. Users can still select **Fixed (single series)** when they want the previous behavior.

# Screenshots

Before the chart defaults to:
Screenshot 2026-07-24 at 7 55 25 AM

After the pr, chart defaults to: 
Screenshot 2026-07-24 at 7 55 43 AM

# Checklist

## UI Changes


Nexucis (approved):
thank you @mixturesolutions !

PR #754, opened 2026-07-24
[FEATURE] Scope TimeSeriesChart annotations to the panel

nicolastakashi:
## What this does

This makes TimeSeriesChart honor panel-level annotations. `TimeSeriesChartPanel` now reads annotations through the new `usePanelAnnotationsWithData` hook, fed the panel's config from `props.definition.spec.annotations`, instead of the global `useAnnotationsWithData`.

The effect: a panel can turn dashboard annotations off for itself, and it renders any annotations defined locally on the panel. The conversion (`convertAnnotationToTimeSeriesAnnotation`) and the chart rendering are untouched, so this is a two-line change to the data source feeding them.

## Dependency

This is stacked on two other PRs:

1. perses/spec#61, which adds `PanelSpec.Annotations` to the data model.
2. The `usePanelAnnotationsWithData` hook in perses/shared.

It will not compile until both land and `@perses-dev/spec` and `@perses-dev/dashboards` are bumped here.

## Testing

The behavior of the hook itself is tested in the perses/shared PR. I have not stood up the full plugins build to compile this in isolation, since it needs the spec, dashboards, and plugin-system packages from the two PRs above. The change is a straight swap of the annotations source, with the same shape flowing into the existing conversion.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -429,15 +429,11 @@
     },
     "node_modules/@adobe/css-tools": {
       "version": "4.5.0",
-      "resolved": "https://registry.npmjs.org/@adobe/css-tools/-/css-tools-4.5.0.tgz",
-      "integrity": "sha512-6OzddxPio9UiWTCemp4N8cYLV2ZN1ncRnV1cVGtve7dhPOtRkleRyx32GQCYSwDYgaHU3USMm84tNsvKzRCa1Q==",
       "dev": true,
       "license": "MIT"
     },
     "node_modules/@asamuzakjp/css-color": {
       "version": "3.2.0",
-      "resolved": "https://registry.npmjs.org/@asamuzakjp/css-color/-/css-color-3.2.0.tgz",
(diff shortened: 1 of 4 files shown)


Gladorme (approved):
lgtm

Gladorme (commented):
Need package.json update before on all plugin for perses deps

package-lock.json
```diff
@@ -642,8 +642,6 @@
     },
     "node_modules/@atlaskit/pragmatic-drag-and-drop-hitbox": {
       "version": "1.2.0",
-      "resolved": "https://registry.npmjs.org/@atlaskit/pragmatic-drag-and-drop-hitbox/-/pragmatic-drag-and-drop-hitbox-1.2.0.tgz",
```
Nexucis, line 645: this diff looks very strange. I don't think the package-lock.json should have been modified since the update of the libs have been done in a different PR

PR #755, opened 2026-07-26
[BUGFIX] Table: align filter rows with table columns 

Aln999:
# Description

This PR fixes the filter row alignment issue when the table is scrolled horizontally. It keeps the filter row aligned with the table columns, ensures the filter widths stay in sync, and fixes the filter search so it matches the displayed values correctly

### Changes

* Added a filter row above the table to filter data column by column.
* Kept the filter row aligned with the table while scrolling horizontally.
* Fixed filter column widths so they always match the table columns.
* Improved filter dropdown positioning so it opens correctly without getting cut off.
* Made the filter dropdown close when the same filter button is clicked again.
* Fixed search inside the filter dropdown to search the displayed values correctly.
* Updated the tests and added the required setup for the new filter row functionality.

# Screenshots

https://github.com/user-attachments/assets/09bc1d17-d763-42b7-8f37-c0768f5e8151

# Checklist

## UI Changes

#Relates to https://github.com/perses/perses/issues/3835

### DIFF
--- a/table/src/components/TablePanel.tsx
+++ b/table/src/components/TablePanel.tsx
@@ -34,6 +34,7 @@ import {
 import { ColumnFiltersState, PaginationState, RowSelectionState, SortingState } from '@tanstack/react-table';
 import { ReactElement, useCallback, useEffect, useMemo, useRef, useState } from 'react';
 import { QueryDataType, TimeSeriesData } from '@perses-dev/spec';
+import { createPortal } from 'react-dom';
 import { CellSettings, ColumnSettings, evaluateConditionalFormatting, TableOptions } from '../models';
 import { buildRawTableData, getTablePanelQueryMode } from '../table-data-utils';
 import { EmbeddedPanel } from './EmbeddedPanel';
@@ -220,7 +221,7 @@ function ColumnFilterDropdown({
   const [searchTerm, setSearchTerm] = useState('');
   const values = [...new Set(allValues)].filter((v) => v !== null).sort();
   const filteredValues = searchTerm
-    ? values.filter((v) => String(v).toLowerCase().includes(searchTerm.toLowerCase()))
+    ? values.filter((v) => String(v.formatted).toLowerCase().includes(searchTerm.toLowerCase()))
     : values;
   if (values.length === 0) {
     return (
@@ -401,6 +402,10 @@ export function TablePanel({ contentDimensions, spec, queryResults }: TableProps
   });
 
   const filteredDataRef = useRef<Array<Record<string, unknown>>>([]);
+  // Refs used to keep the filter row in sync with the table's horizontal
+  const panelContainerRef = useRef<HTMLDivElement>(null);
+  const filterRowInnerRef = useRef<HTMLDivElement>(null);
+  const filterCellRefs = useRef<Array<HTMLDivElement | null>>([]);
 
   // Convert selectionMap to TanStack's RowSelectionState format
   const rowSelection = useMemo((): RowSelectionState => {
@@ -682,6 +687,12 @@ export function TablePanel({ contentDimensions, spec, queryResults }: TableProps
   const handleFilterClick = (event: React.MouseEvent<HTMLButtonElement>, columnId: string): void => {
     event.preventDefault();
     event.stopPropagation();
+
+    if (openFilterColumn === columnId) {
+      handleFilterClose();
+      return;
+    }
+
     setFilterAnchorEl({ ...filterAnchorEl, [columnId]: event.currentTarget });
     setOpenFilterColumn(columnId);
   };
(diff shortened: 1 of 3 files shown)


table/src/components/TablePanel.tsx
```diff
@@ -728,6 +733,77 @@ export function TablePanel({ contentDimensions, spec, queryResults }: TableProps
...
+    const filterRowInner = filterRowInnerRef.current;
+
+    if (!scrollContainer || !filterRowInner) {
+      return;
+    }
+
+    const syncFilterRowScroll = (): void => {
+      filterRowInner.style.transform = `translateX(-${scrollContainer.scrollLeft}px)`;
+
+      setOpenFilterColumn((current) => (current === null ? current : null));
```
shahrokni, line 758: It seems the set value is `null ` anyway. 
Your logic => if current is `null` set it to current, otherwise set it to `null`.
So, couldn't you simply do something like this? Am I missing something?

```typeScript
setOpenFilterColumn(null);
```

table/src/components/TablePanel.tsx
```diff
@@ -728,6 +733,77 @@ export function TablePanel({ contentDimensions, spec, queryResults }: TableProps
...
+        filterCell.style.width = width;
+        filterCell.style.minWidth = width;
+        filterCell.style.maxWidth = width;
+      });
+    };
+
+    syncColumnWidths();
+
+    // Re-sync whenever the header row's size changes 
+    const headerRow = scrollContainer.querySelector('thead tr');
```
shahrokni: You could've lifted this line up next to the 

```typescript
const scrollContainer = panelContainerRef.current?.querySelector<HTMLElement>('.MuiTableContainer-root');
const headerRow = scrollContainer?.querySelector('thead tr');

if(!headerRow){
   return;
}
```
With an early return you could've avoided the execution of some lines and also unnecessary heap memory consumption. WDYT? But before taking any action, does it make sense to initially call `syncColumnWidths();` even if headerRow is undefined? Because, if this is the case, my suggestion does not make any sense.

table/src/components/TablePanel.tsx
```diff
@@ -748,99 +824,119 @@ export function TablePanel({ contentDimensions, spec, queryResults }: TableProps
...
+            }}
+          >
+            {columns.map((column, idx) => {
+              const filters = getSelectedFilterValues(column.accessorKey as string);
+              const columnWidth = column.width || spec.defaultColumnWidth;
+              const anchorEl = filterAnchorEl[column.accessorKey as string];
+
+              return (
+                <div
+                  key={`filter-${idx}`}
```
shahrokni: I know this is the legacy part. But let's see if we can improve it. 
Although the code is looping over the `columns` the return element key has been named `filter-${idx}`
Probably the key should've included `column-${something}`
Now, lets say my assumption is correct (we need to check that first)
Then the key is not really performant. According to official React docs, assigning an index-based key is not a good idea. So, couldn't the original developer utilize `column.accessorKey`, it seems it is unique. Let's take a look and see if we can improve this key carefully. (Make sure `accessorKey` is unique)

table/src/components/TablePanel.tsx
```diff
@@ -748,99 +824,119 @@ export function TablePanel({ contentDimensions, spec, queryResults }: TableProps
...
+
+                  {openFilterColumn === column.accessorKey &&
+                    anchorEl &&
+                    createPortal(
+                      <div
+                        style={{
+                          position: 'fixed',
+                          top: anchorEl.getBoundingClientRect().bottom + 4,
+                          left: anchorEl.getBoundingClientRect().left,
+                          zIndex: 1300,
```
shahrokni: Where does this `1300` come from? Any logic or specific reason?

shahrokni:
@Aln999 
You also need to `rebase main`

> This branch is out-of-date with the base branch
> Merge the latest changes from main into this branch

Please let me know if you need help for that

shahrokni:
@Aln999 
You have failing pipelines. Please take a look and resolve the issues. The issues are tiny. 
Since, the pipeline is not automatically running automatically after every push, you can check the linters, tests, etc locally to make sure that everything is in order before pushing to the remote side.

shahrokni:
@Aln999 
There is an old bug here and I am wondering, if this is the best time to get rid of it! However, if it only takes less than an hour! 
So, feel free to ignore this one, if it demands some times and effort. 
You see the filter button here? The click wouldn't close the filter. Could we fix this?

image

shahrokni:
I tested the table PR and your fix amazingly have made it way better! I really appreciate for the contribution. I tried to be picky and find something to complain about but I couldn't! Thank you @Aln999 🚀

table/src/components/TablePanel.tsx
Aln999: no good reason to hard-code it. It happens to match MUI's own theme.zIndex.modal so I've referenced that directly instead of the magic number. Will update it to
```typescript
zIndex: theme.zIndex.modal,
```

table/src/components/TablePanel.tsx
Aln999: You're right, it doesn't make sense to run syncColumnWidths() without a header row. Rearranged to return early

table/src/components/TablePanel.tsx
Aln999: This follows the lint rule because it uses the current state value before changing it. It also checks if the value is already null and skips the update in that case, avoiding an unnecessary state change.

table/src/components/TablePanel.tsx
```diff
@@ -748,99 +833,118 @@ export function TablePanel({ contentDimensions, spec, queryResults }: TableProps
...
+            }}
+          >
+            {columns.map((column, idx) => {
+              const filters = getSelectedFilterValues(column.accessorKey as string);
+              const columnWidth = column.width || spec.defaultColumnWidth;
+              const anchorEl = filterAnchorEl[column.accessorKey as string];
+
+              return (
+                <div
+                  key={`filter-${column.accessorKey}`}
```
shahrokni: As we discussed should be 

```javaScript
key={`column-${column.accessorKey}`}
```
No?

shahrokni (approved):

PR #524, opened 2026-01-16
[FEATURE] enhance StatChart with text mode options and refactor display l…

abelyakin:
# Description

following issue #501 
Implements Grafana's text mode feature to provide users with explicit control over what's displayed in stat panels. Previously, users were confused about when to use metricLabel vs legendFormat, and there was no way to show series names alongside values.

legendFormat: Template for formatting series names (e.g., {{instance}} - {{job}}). Applied to all series for display purposes.
metricLabel: Selects which label's value to display instead of the calculated number (e.g., show the "status" label value like "healthy" or "degraded" rather than a numeric metric).
textMode: Controls what to show - the calculated value, series name, both, or neither.

Updated and simplified the migration logic to map `textMode` from panel options and handle various display scenarios. 
Grafana's textMode values pass through directly without conversion. Existing panels default to auto for backward compatibility.

# Screenshots
### text mode = name
Screenshot 2026-01-16 at 03 29 07
### text mode = value
Screenshot 2026-01-16 at 03 29 18
### text mode = value and name
Screenshot 2026-01-16 at 03 29 40

# Checklist

- [x] Pull request has a descriptive title and context useful to a reviewer.
- [x] Pull request title follows the `[<catalog_entry>] <commit message>` naming convention using one of the
  following `catalog_entry` values: `FEATURE`, `ENHANCEMENT`, `BUGFIX`, `BREAKINGCHANGE`, `DOC`,`IGNORE`.
- [x] All commits have [DCO signoffs](https://github.com/probot/dco#how-it-works).

## UI Changes

- [x] Changes that impact the UI include screenshots and/or screencasts of the relevant changes.
- [x] Code follows the [UI guidelines](https://github.com/perses/perses/blob/main/ui/ui-guidelines.md).

PR #758, opened 2026-07-28
[FEATURE] enhance StatChart with text mode options and refactor display logic

Nexucis:
this PR is the same as https://github.com/perses/plugins/pull/524. I am just solving the merge conflict and I have preserved the commit from @abelyakin.

I am opening this PR in order to be able to merge it as the merge conflict was the last thing to solve before merging it.


Nexucis:
the cue diff does not make sense for me ... I will check later what's going one here

statchart/schemas/migrate/migrate.cue
```diff
@@ -37,7 +37,7 @@ spec: {
 	calculation: *commonMigrate.#mapping.calc[#panel.options.reduceOptions.calcs[0]] | commonMigrate.#defaultCalc // only consider [0] here as Perses's GaugeChart doesn't support individual calcs
 
 	// metricLabel
-	#textMode: *#panel.options.textMode | null
+	#textMode: *#panel.options.textMode | "auto"
```
SoWieMarkus, line 40: @Nexucis regarding the cue issue.

I checked the history of this code snippet :D looks like the `textMode: #textMode` output line got dropped while resolving merge conflicts. So the migration was computing the value but never actually writing it to the output.

I tested it locally with that change :) However there is also a `"textMode": "auto"` missing in `schemas/migrate/testspecial_mapping_without_match` which needs to be added separately.

```suggestion
	#textMode: *#panel.options.textMode | "auto"
	textMode: #textMode
```

Nexucis: sorry @SoWieMarkus I went to something else. Thank you for helping me with this PR ! I will try it and see if there is other issue after applying your suggestion.

Nexucis:
mmm looking at the conflict, I think there is even a conflict in the logic of managing the legend here.

I will close this PR as I am not sure how to merge these logics together. That's a pity because the feature is still interesting :/

PR #773, opened 2026-08-19
[BUGFIX] rsbuild: handle plugin version

Gladorme:
# Description

In spec of plugin we can define "metadata" (version + registry) and when I wanted to implement it on frontside, I got some issue when there is multiple instances of a same plugin with different versions.

Makes plugin builds version-aware so multiple versions of the same plugin can be installed and loaded side by side. Previously only the latest installed version of a plugin was actually loadable — any other version either failed to load or silently rendered the latest one's code.

**1. Version in the asset prefix**

  `getPublicPath` now returns `/plugins/<name>~<version>/` instead of `/plugins/<name>/`. The version is read from the plugin's `package.json` at build time.

  **Why:** a manifest fetched from `/plugins/<name>~<version>/mf-manifest.json` resolved its (relative) remoteEntry against the version-less prefix. The server maps a path with no `~version` to  `LatestVersion`, so an older version requested the latest version's directory. Since chunk filenames are content-hashed, the file didn't exist there:

  `RUNTIME-008 ScriptNetworkError resourceUrl: /plugins/TimeSeriesChart/__mf/js/TimeSeriesChart.7c5da441.js (404)`

  This also explains why the newest version always appeared to work, and why an older version worked right up until a newer archive was dropped in.

  **2. Version in the Module Federation container global name**

  Sets library: `{ type: 'global', name: '<Name>_<version>' }`, so the manifest's globalName is unique per version.

  **Why:** MF resolves a remote's container via `globalThis[globalName]`, taken from the manifest (`assignRemoteInfo` in `@module-federation/runtime-core`), and `loadEntryScript` early-returns an already-registered container:

```
  const { entryExports } = getRemoteEntryExports(name, globalName);
  if (entryExports) return entryExports;
```

  With every version sharing the global name, the first version loaded won and later versions silently reused its container without ever fetching their own entry.

  **3. Version in the webpack chunk registry**

  Sets `output.uniqueName` and `output.chunkLoadingGlobal` to `chunk_<Name>_<version>`.

  **Why:** async chunks register into a shared global array:

```
  (self.chunk_TimeSeriesChart = self.chunk_TimeSeriesChart || []).push([["53"], { 10985(...) }])
```

  When two versions share that global, the second one pushes its chunks into the runtime the first installed, and because module IDs are deterministic they collide: both versions resolve to the first one's modules. `chunkLoadingGlobal` is set explicitly because rspack derives it from `uniqueName` before the tools.rspack hook runs, so setting `uniqueName` alone is not enough.

**Plugins must be rebuilt for this to take effect; it is not retroactive for already-published archives**
I did not handle registry name to have super long name, because I am not sure it will happen often 🙏 
I guess if someone has the issue, he can change plugin name as quick win/fix

# Checklist

## UI Changes

### DIFF
--- a/rsbuild.shared.ts
+++ b/rsbuild.shared.ts
@@ -11,6 +11,9 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import { readFileSync } from 'node:fs';
+import { resolve } from 'node:path';
+
 import { ModuleFederationOptions, pluginModuleFederation } from '@module-federation/rsbuild-plugin';
 import { mergeRsbuildConfig, RsbuildConfig } from '@rsbuild/core';
 
@@ -73,7 +76,7 @@ interface PluginConfigOptions {
  * });
  * ```
  */
-export function createConfigForPlugin(options: PluginConfigOptions) {
+export function createConfigForPlugin(options: PluginConfigOptions): RsbuildConfig {
   const { name, rsbuildConfig = {}, moduleFederation = {} } = options;
 
   const mfConfig: ModuleFederationOptions = {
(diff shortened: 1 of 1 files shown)


rsbuild.shared.ts
```diff
@@ -131,17 +153,55 @@ function getRsbuildConfig(name: string): RsbuildConfig {
         if (process.env.NODE_ENV !== 'development') {
           config.output.publicPath = 'auto';
         }
+        // Isolate each version's webpack runtime.
```
jgbernalp, line 156: this probably is not needed as we should use the path, regardless if the chunk has the same name

ibakshay:
Thank you for your work. It works great already. I just found a small bug when I was testing. 
Both `lock` and `unlock` button shows up after you edit a panel by changing the description or query. And the version  property in the panel gets removed when you apply the changes. 

https://github.com/user-attachments/assets/6cf480e8-2623-4993-a92d-689b5f25f9e4

rsbuild.shared.ts
Gladorme: Pretty sure if I remove it, it will fail :/
I will check tomorrow

Gladorme:
> Thank you for your work. It works great already. I just found a small bug when I was testing. Both `lock` and `unlock` button shows up after you edit a panel by changing the description or query. And the version property in the panel gets removed when you apply the changes.
> 
>  Screen.Recording.2026-08-22.at.21.58.22.mov

Issues fixed: https://github.com/perses/spec/pull/74 and for double buttons 😄

rsbuild.shared.ts
Gladorme: I confirm without this unique identifier, there is issue (it's only using chunks from higher version). How to replicate in local: hardcode a color for series in timeserieschart panel (`TimeSeriesChartPanel.tsx`), update plugin version, build plugin (`percli plugin build --plugin.path=./timeserieschart --skip.npm-install=true`). Redo this a second time with another color and other plugin version.
Then check in panel editor preview when changing version in panel type selection, it will display the same color :/


Color example:
```
const seriesColor = `hsl(130 100% 50%)`;
const seriesColor = `hsl(276 100% 50%)`;
```

jgbernalp:
I'm checking from my side but other plugins different from prometheus still don't work. It seems the backend is not loading the latest version when no registry or version are provided. I'm testing...

jgbernalp:
I saw an issue with the backend version check, fixing it with: https://github.com/perses/perses/pull/4387

I'm still checking from the frontend side...

rsbuild.shared.ts
jgbernalp: ok looking this in detail, I understand now this happens when two exact same versions run in parallel, for example in dev mode.

jgbernalp (approved):

PR #774, opened 2026-08-20
[FEATURE] BarChart: add regex based color override

SoWieMarkus:
# Description

Adds support for regex-based color overrides in the BarChart panel. Previously users had no way to control this. 

With this change, users can define a list of `{ regex, color }` overrides in the panel settings. A color is applied when a regex matches, otherwise the chart falls back to the palette as before.

Behavior details:
- The regex is matched against **both** series names (grouped/stacked) **and** bar labels (ungrouped).
- When multiple overrides match, the **first** matching entry in the list wins.
- Empty and invalid regex patterns are ignored (skipped) rather than throwing an error or matching all values.

# Screenshots

### Without stacking, matches multiple bars
Bildschirmfoto 2026-08-20 um 11 12 09
Bildschirmfoto 2026-08-20 um 11 12 01

### Without stacking, matches all bars

Bildschirmfoto 2026-08-20 um 11 36 37
Bildschirmfoto 2026-08-20 um 11 36 32

### With stacking

Bildschirmfoto 2026-08-20 um 11 39 11
Bildschirmfoto 2026-08-20 um 11 39 17

### General updated settings view

With overrides:
Bildschirmfoto 2026-08-20 um 11 51 03

Empty:
Bildschirmfoto 2026-08-20 um 11 51 14

# Checklist

- [x] Pull request title follows the `[<catalog_entry>] <commit message>` naming convention using one of the following `catalog_entry` values: `FEATURE`, `ENHANCEMENT`, `BUGFIX`, `BREAKINGCHANGE`, `DOC`,`IGNORE`.

## UI Changes

### DIFF
--- a/barchart/schemas/tests/valid/bar_color_overrides.json
+++ b/barchart/schemas/tests/valid/bar_color_overrides.json
@@ -7,11 +7,13 @@
     },
     "sort": "desc",
     "mode": "value",
-    "colorOverrides": [
-      {
-        "regex": "errors",
-        "color": "#ff0000"
-      }
-    ]
+    "visual": {
+      "colorOverrides": [
+        {
+          "regex": "errors",
+          "color": "#ff0000"
+        }
+      ]
+    }
   }
 }
--- a/barchart/schemas/tests/valid/bar_color_overrides.json
+++ b/barchart/schemas/tests/valid/bar_color_overrides.json
@@ -1,0 +1,17 @@
+{
+  "kind": "BarChart",
+  "spec": {
+    "calculation": "last-number",
+    "format": {
+      "unit": "decimal"
+    },
+    "sort": "desc",
+    "mode": "value",
+    "colorOverrides": [
+      {
+        "regex": "errors",
+        "color": "#ff0000"
+      }
+    ]
+  }
+}
--- a/barchart/src/BarChartBase.tsx
+++ b/barchart/src/BarChartBase.tsx
@@ -26,6 +26,9 @@ import { use as registerECharts, EChartsCoreOption } from 'echarts/core';
 import { CanvasRenderer } from 'echarts/renderers';
 import { ReactElement, useMemo } from 'react';
 
+import { ColorOverride } from './bar-chart-model';
+import { getOverrideColor } from './utils';
+
 registerECharts([
   EChartsBarChart,
   GridComponent,
@@ -64,6 +67,7 @@ export interface BarChartBaseProps {
   groupedData?: StackedBarChartData | null;
   isStacked?: boolean;
   orientation?: 'horizontal' | 'vertical';
+  colorOverrides?: ColorOverride[];
 }
 
 export function BarChartBase(props: BarChartBaseProps): ReactElement {
@@ -76,6 +80,7 @@ export function BarChartBase(props: BarChartBaseProps): ReactElement {
     groupedData,
     isStacked = false,
     orientation = 'horizontal',
+    colorOverrides,
   } = props;
   const chartsTheme = useChartsTheme();
   const isHorizontal = orientation === 'horizontal';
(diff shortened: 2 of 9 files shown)


barchart/schemas/tests/valid/bar_color_overrides.json
```diff
@@ -0,0 +1,17 @@
+{
+  "kind": "BarChart",
+  "spec": {
+    "calculation": "last-number",
+    "format": {
+      "unit": "decimal"
+    },
+    "sort": "desc",
+    "mode": "value",
+    "colorOverrides": [
```
jgbernalp: We usually group these options into a `visual` property:

```suggestion
    "visual": {
       "colorOverrides": [
```

jgbernalp (approved):
LGTM, what do you think @Gladorme ?

Gladorme:
Look nice 😄 I will test quickly. I think we will want to add threshold with color too in the future. I wonder how we decide who take priority. I think we have a similar issue on table panel for cell color 🤔

barchart/src/BarChartOptionsEditorSettings.tsx
```diff
@@ -184,6 +230,30 @@ export function BarChartOptionsEditorSettings(props: BarChartOptionsEditorProps)
             label="Stack bars"
           />
         </OptionsEditorGroup>
+        <OptionsEditorGroup title="Color Overrides">
```
Gladorme, line 235: Please move it in a new column `<OptionsEditorColumn>`, there is space available 😄

barchart/src/BarChartBase.tsx
```diff
@@ -182,7 +193,12 @@ export function BarChartBase(props: BarChartBaseProps): ReactElement {
         },
         itemStyle: {
           borderRadius: 4,
-          color: chartsTheme.echartsTheme[0],
+          color: (params: { name: string }): string => {
+            const overrideColor = getOverrideColor(params.name, colorOverrides);
+            if (overrideColor) return overrideColor;
+            const palette = chartsTheme.echartsTheme.color as string[] | undefined;
+            return palette?.[0] ?? '';
```
Gladorme, line 200: I would said:
```suggestion
            return chartsTheme.echartsTheme[0];
```

Based on previous code, but maybe I miss something

barchart/src/bar-chart-model.ts
```diff
@@ -21,6 +21,23 @@ export const DEFAULT_MODE: ModeOption = 'value';
 export const DEFAULT_ORIENTATION: 'horizontal' | 'vertical' = 'horizontal';
 export const DEFAULT_GROUP_BY: string[] = [];
 export const DEFAULT_IS_STACKED = false;
+export const DEFAULT_COLOR_OVERRIDES: ColorOverride[] = [];
```
Gladorme: Remove, not used

SoWieMarkus: Good catch, thanks!

barchart/src/BarChartOptionsEditorSettings.tsx
SoWieMarkus: Moved it :)

barchart/src/BarChartBase.tsx
SoWieMarkus: I had this first, but without my change the default setting chart bars didn't have any color by default. 
I checked the type and since echartsTheme is an object and not an array i don't know what the previous tried to do - and why it worked :D Maybe you can help me understand that? Have you tried it locally?

Gladorme: No I did not tested this change locally, but I tested your PR and it's worked, I guess it's fine

Gladorme (approved):
lgtm, thanks :)

PR #781, opened 2026-08-26
feat(statchart): use an instant query when the value allows it

s3onghyun:
## What

StatChart always issued a range query, even when it only needed the latest point. Now that #752 added a per-query instant mode, this derives the panel query mode from the spec — the same thing `TablePanel` already does via `getTablePanelQueryMode`.

`getStatChartQueryMode` returns `instant` only when it's safe:
- **`last`** → instant (it's exactly what an instant query returns)
- **`last-number`, `mean`/`sum`/`min`/`max`, `first`/`first-number`** → range (they need the full series)
- **any spec with a sparkline** → range (it draws the series)

So the default (sparkline on) stays range, and nothing that renders the series changes — the #738 sparkline regression is kept covered by tests.

Follow-up to #746, which was closed in favour of #752's per-query flag.

## Test

`stat-chart-model.test.ts` covers each calculation + the sparkline cases. `npm run test`, `tsc`, and `oxlint` are clean.

### DIFF
--- a/statchart/src/StatChart.ts
+++ b/statchart/src/StatChart.ts
@@ -13,7 +13,7 @@
 
 import { PanelPlugin } from '@perses-dev/plugin-system';
 
-import { createInitialStatChartOptions, StatChartOptions } from './stat-chart-model';
+import { createInitialStatChartOptions, getStatChartQueryOptions, StatChartOptions } from './stat-chart-model';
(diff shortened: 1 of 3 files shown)


s3onghyun:
Rebased onto `main` (the conflict was just the new `import type` split) — tests and `tsc` are green. Ready for review whenever you have a moment. @Gladorme

PR #800, opened 2026-09-07
[FEATURE] TimeSeriesChart: add Exemplars support

julianocosta89:
# Description

Part of #3445 (perses/perses#3445). **Stacked on** https://github.com/perses/plugins/pull/799 — review/merge that one first. Also uses `ExemplarMetadataDialog` from perses/shared     
 (<shared PR link>).                                                                                                                                                                     
                                                                                                                                                                                         
   ### Summary                                                                                                                                                                           
                                                                                                                                                                                         
   Renders the exemplar data fetched by the Prometheus datasource changes as **diamond scatter series** on time-series charts, with a metadata dialog on click — Grafana-style, as       
 discussed in #3445.                                                                                                                                                                     
                                                                                                                                                                                         
   ### Changes                                                                                                                                                                           
                                                                                                                                                                                         
   - Exemplars render as diamond scatter series on top of their matching series (ECharts supports mixing series types in one chart), inheriting the series color and y axis.             
   - Only rendered for series visible in the legend — exemplar series are dropped for deselected series.                                                                                 
   - **Clicking a diamond opens the `ExemplarMetadataDialog`** from `@perses-dev/components` (series labels, exemplar labels, value, timestamp) instead of pinning the chart tooltip.    
   - Minor touch-ups in `prometheus/` carried over from the stacked review (redundant type re-exports removed).                                                                          
                                                                                                                                                                                         
   ### Out of scope                                                                                                                                                                      
                                                                                                                                                                                         
   Trace-linking in the Traces explorer, HeatMapChart rendering (can reuse the shared dialog), hover-tooltip integration — as discussed in #3445.                                        
                                                                                                                                                                                         
   ### Testing                                                                                                                                                                           
                                                                                                                                                                                         
   - TimeSeriesChart: 62 tests pass, including panel-level tests asserting exemplars reach the ECharts option as diamond scatter series with embedded metadata, and are dropped for      
 deselected series (option captured via a mocked `EChart` component since jsdom has no canvas).                                                                                          
   - Repo-wide: `turbo run type-check` / `test` (30 packages), `lint` (31), `oxfmt` clean.

# Screenshots

See https://github.com/perses/shared/pull/278

# Checklist

## UI Changes


AntoineThebaud:
Please add a screenshot to your PR desc 🙏

julianocosta89:
Addressed all 6 Copilot review comments:
   1. negativeY exemplars — fixed in data-transform.ts + panel + hover tracking (unit test added)
   2. global.fetch leak — vi.stubGlobal + afterEach unstubAllGlobals
   3. side effects in updater — compute state first, call setters separately
   4. duplicate no-assertion test — removed
   5. userEvent.click await — N/A, repo pins user-event v13 (sync API, no setup())
   6. toSorted — N/A, repo lint rule unicorn/no-array-sort mandates toSorted

   Signed-off-by: Juliano Costa <juliano.costa@datadoghq.com>

timeserieschart/src/TimeSeriesChartPanel.tsx
```diff
@@ -302,6 +327,17 @@ export function TimeSeriesChartPanel(props: TimeSeriesChartProps): ReactElement
...
+            if (timeSeries.labels) {
+              seriesByLabels.set(labelsKey(timeSeries.labels), {
+                seriesId,
+                seriesName: formattedSeriesName,
+                color: seriesColor,
+                seriesLabels: timeSeries.labels,
+                yAxisIndex,
+                exemplars: [],
+              });
+            }
```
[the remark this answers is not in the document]
julianocosta89: Addressed as suggested: `ExemplarChartData` now carries a `negativeY` flag inherited from the query's `querySettings`, and  `getExemplarSeries` applies it to the plotted Y value (`value: [timestamp, -value]`) while `exemplar.value` keeps the original value for the metadata tooltip. I also updated the hover tracking in `TimeSeriesChartBase` to use the plotted value, so `convertToPixel` still finds negated markers for the mouse-follow tooltip. Covered by a new unit test in `data-transform.test.ts`. Fixed in 0d4db23d.

timeserieschart/src/TimeSeriesChartPanel.test.tsx
```diff
@@ -92,14 +108,77 @@ describe('TimeSeriesChartPanel', () => {
...
+      expect(getExemplarSeries()).toHaveLength(0);
+    });
+
+    it('should only render exemplars of series selected in the legend', async () => {
+      renderPanel({ ...MOCK_TIME_SERIES_DATA_MULTIVALUE, exemplars: MOCK_TIME_SERIES_EXEMPLARS });
+      await waitFor(() => {
+        expect(getExemplarSeries()).toHaveLength(2);
+      });
+
+      userEvent.click(getLegendByName(MOCK_TIME_SERIES_DATA_MULTIVALUE.series[0]?.name));
```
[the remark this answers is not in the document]
julianocosta89: Not applicable: the repo pins `@testing-library/user-event` v13 (13.5.0), whose direct `userEvent.click()` API is synchronous — the v14 `setup()` / `await user.click()` API doesn't exist in this version (switching would fail type-check). The sync click followed by `waitFor` is the correct pattern for v13, and matches the existing legend tests in this file. I've added a NOTE comment in the test documenting this. See 0d4db23d.

timeserieschart/src/TimeSeriesChartPanel.tsx
```diff
@@ -65,6 +66,18 @@ import { getSeriesColor } from './utils/palette-gen';
...
 export type TimeSeriesChartProps = PanelProps<TimeSeriesChartOptions, TimeSeriesData>;
 
+/**
+ * Stable string key for a labels record, so exemplars can be matched to their
+ * series regardless of the labels order.
+ */
+function labelsKey(labels: Labels): string {
+  return JSON.stringify(
+    Object.keys(labels)
+      .toSorted()
```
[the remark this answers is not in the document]
julianocosta89: Not applicable: the repo's Oxlint config enforces `unicorn/no-array-sort`, which requires `Array#toSorted()` and flags `Array#sort()` — switching would add a new lint warning, which the contributor guidelines forbid. `toSorted()` is already used elsewhere in this codebase (e.g. `prometheus/src/utils/utils.ts`, `QuerySettingsEditor.tsx`), so the TS/JS target already guarantees runtime support. Keeping as is. See 0d4db23d.

julianocosta89:
@AntoineThebaud this is the last piece of the puzzle 🥳 
Would you be able to take a look?

AntoineThebaud:
> @AntoineThebaud this is the last piece of the puzzle 🥳 Would you be able to take a look?

Yes @julianocosta89 sorry I'm busy with several topics, but for sure I'll take time this week to try the whole MVP locally & drop a review

timeserieschart/src/TimeSeriesChartPanel.tsx
AntoineThebaud, line 76: When trying locally with a basic query `rate(traces_spanmetrics_latency_bucket[5m])` I wasn't able to display any exemplar in the first place, and it looks like the logic was missing this. Can you confirm?

```suggestion
      // Prometheus includes `__name__` in /query_exemplars seriesLabels but omits
      // it from /query_range metric labels. Thus we ignore it otherwise the series never matches correctly.
      .filter((labelName) => labelName !== '__name__')
      .toSorted()
```

timeserieschart/src/TimeSeriesChartPanel.test.tsx
```diff
@@ -92,14 +108,75 @@ describe('TimeSeriesChartPanel', () => {
...
+      }
+
+      // eslint-disable-next-line @typescript-eslint/no-explicit-any
+      const vda1Series = exemplarSeries.find((s) => (s as any)?.id?.includes('vda1')) as any;
+      const firstItem = vda1Series?.data?.[0];
+      const expectedExemplars = MOCK_TIME_SERIES_EXEMPLARS[0]?.exemplars ?? [];
+      expect(firstItem?.exemplar).toEqual(expectedExemplars[0]);
+      expect(firstItem?.seriesLabels).toEqual(MOCK_TIME_SERIES_EXEMPLARS[0]?.seriesLabels);
+    });
+
```
AntoineThebaud, line 152: If the issue spotted [here](https://github.com/perses/plugins/pull/800/changes#r4080724799) is confirmed, add test to cover it:

```suggestion
    it('should match exemplars whose Prometheus labels include __name__', async () => {
      renderPanel({
        ...MOCK_TIME_SERIES_DATA_MULTIVALUE,
        exemplars: MOCK_TIME_SERIES_EXEMPLARS.map((series) => ({
          ...series,
          seriesLabels: { ...series.seriesLabels, __name__: 'node_filesystem_avail_bytes' },
        })),
      });

      await waitFor(() => {
        expect(getExemplarSeries()).toHaveLength(2);
      });
    });
```

AntoineThebaud (commented):
Sorry for the delay on this @julianocosta89! Please have a look at the problem I stumbled upon & confirm whether this is an expected behavior or something to fix?

timeserieschart/src/TimeSeriesChartPanel.tsx
julianocosta89: Confirmed — great catch! Verified against a live Prometheus: `/query_exemplars` returns `seriesLabels` with `__name__` of the underlying series, while `/query_range` drops `__name__` for function expressions like `rate(...)`, so the labels key never matched and exemplars were silently dropped. Applied your suggestion in 4f5227df.

timeserieschart/src/TimeSeriesChartPanel.test.tsx
julianocosta89: Done — added the test in 4f5227df (slightly reworked with `Object.assign` instead of spreads in the `map`, because the repo's Oxlint config flags `no-map-spread` and new warnings are forbidden). It renders exemplars whose `seriesLabels` include `__name__` and asserts both exemplar series are produced.

timeserieschart/src/TimeSeriesChartPanel.tsx
AntoineThebaud: Sorry I just realized this is actually not where such fix should live, as we want to keep TimeSeriesChart & exemplars rendering logic agnostic from the datasource that was used in the first place. It should thus be a fix on Prometheus plugin side to filter the data before passing it to the chart.

AntoineThebaud: Could you thus revert this last change? We can then merge this PR & do a follow-up for this prom adjustment

julianocosta89: Reverted in 2383ed51 (the `__name__` filter in `TimeSeriesChartPanel.tsx` plus the accompanying panel test, restoring the pre-4f5227df state). Keeping TimeSeriesChart datasource-agnostic as suggested; I'll open the follow-up PR on the Prometheus plugin side to strip `__name__` from `seriesLabels` in the exemplar transformation before it reaches the chart.

timeserieschart/src/TimeSeriesChartPanel.test.tsx
julianocosta89: Reverted in 2383ed51 along with the `__name__` filter it covered — the matching fix (and this test, adapted to the Prometheus plugin's transform) will land in the follow-up PR on the Prometheus plugin side, per the same thread above.

AntoineThebaud (approved):
Awesome! Thanks @julianocosta89 for the hard work 🚀

timeserieschart/src/TimeSeriesChartPanel.tsx
AntoineThebaud: Thanks, I will do the followup for the fix

PR #803, opened 2026-09-08
[BUGFIX] prometheus: evaluate instant queries at the exact end of the time range

rajusem:
# Description

`getTimeSeriesData` aligns the start and end of the time range to the step before sending the query. Range queries need that (stable samples, x-axis), but instant queries have no step: aligning their evaluation time moves it back by up to one step. The step depends on the time range *and the panel width*, so on a 1-week range a full-width Table panel evaluates ~5 min behind "now" and a narrower panel with the same query up to 30 min behind, and two panels on the same dashboard can evaluate the same query at different instants.

Observed on an OpenShift console dashboard (Perses Table panels in instant mode): after a VM stopped, its rows stayed in the table for up to one step on a 1-week range, while they disappeared immediately on a 1-hour range.

This change only aligns the time range for range queries. Instant queries (`context.mode === 'instant'` or `spec.instant: true`) now use the exact end of the time range, which is what the Loki plugin already does (`get-loki-time-series-data.ts`). The returned `timeRange` follows the same rule. Range queries are unchanged, including the "step larger than range" guard from #213.

Test added: with a range end of `06:07:23` and a 30-min step, the instant query must be sent with `time = 06:07:23` (before this change it was sent with `06:00:00`).

# Screenshots

No UI change.

# Checklist

## UI Changes

- [ ] Changes that impact the UI include screenshots and/or screencasts of the relevant changes. (N/A)
- [ ] Code follows the [UI guidelines](https://github.com/perses/perses/blob/main/ui/ui-guidelines.md). (N/A)

### DIFF
--- a/prometheus/src/plugins/prometheus-time-series-query/get-time-series-data.ts
+++ b/prometheus/src/plugins/prometheus-time-series-query/get-time-series-data.ts
@@ -73,24 +73,33 @@ export const getTimeSeriesData: TimeSeriesQueryPlugin<PrometheusTimeSeriesQueryS
   const timeRange = getPrometheusTimeRange(context.timeRange);
   const step = getRangeStep(timeRange, minStep, undefined, context.suggestedStepMs); // TODO: resolution
 
-  // Align the time range so that it's a multiple of the step
+  // `spec.instant` is a per-query override: `true` forces instant, `false` forces range.
+  // When left unset (Auto), defer to the panel-provided `context.mode`.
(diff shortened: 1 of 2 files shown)


rajusem:
@Nexucis @jgbernalp could you take a look as well? This touches the instant-query path added in #752 (Prometheus plugin) and the behaviour shows up in Table panels in instant mode.

jgbernalp (approved):

PR #642, opened 2026-04-27
[FEATURE] prometheus & timeserieschart: Add annotation support

Gladorme:
# Description

Add annotation generation for Prometheus datasource and add display on TimeSeriesChart.

Idea: add option to ignore "end" to only keep a point

PR depends on: 
- https://github.com/perses/shared/pull/120 (needed for TS, imports)
- https://github.com/perses/perses/pull/4040 (needed for test, percli must support Annotation plugin)

# Screenshots

image

# Checklist

- [ ] Pull request has a descriptive title and context useful to a reviewer.
- [ ] Pull request title follows the `[<catalog_entry>] <commit message>` naming convention using one of the
  following `catalog_entry` values: `FEATURE`, `ENHANCEMENT`, `BUGFIX`, `BREAKINGCHANGE`, `DOC`,`IGNORE`.
- [ ] All commits have [DCO signoffs](https://github.com/probot/dco#how-it-works).

## UI Changes

- [ ] Changes that impact the UI include screenshots and/or screencasts of the relevant changes.
- [ ] Code follows the [UI guidelines](https://github.com/perses/perses/blob/main/ui/ui-guidelines.md).

PR #804, opened 2026-09-08
[DOC] Document the Prometheus annotation plugin

7487:
# Description

The Prometheus package ships an annotation plugin since #642, but `docs/prometheus` still lists only the datasource, the variables, the time series query and the explorer.

- `docs/prometheus/README.md`: add an `Annotation (PrometheusPromQLAnnotation)` section.
- `docs/prometheus/model.md`: add the corresponding data model entry, so the section can link to it like the others do.

The described behaviour comes from `prometheus/src/annotations/get-annotation-data.ts` (range query over the dashboard time range, one annotation per returned series, spanning first to last sample, `{{label_name}}` interpolation in `title`/`legend` via `formatSeriesName`, `tags` restricting the labels shown) and from `prometheus/schemas/prometheus-promql-annotation`.

Part of perses/perses#4408, whose other half (documenting the plugin kind itself) is perses/perses#4427.

One thing I did not touch, but that looks like a typo: `prometheus-promql-annotation.cue` declares `tags?: [string]`, i.e. a list of exactly one element, while `PrometheusPromQLAnnotationOptions.tags` is a `string[]` and the option editor lets you add several. The generator template uses `[...string]`. The example in the docs uses a single tag so that it validates either way. Let me know if you want the schema fixed here or in a separate PR.

# Checklist

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/prometheus/README.md
+++ b/docs/prometheus/README.md
@@ -78,6 +78,15 @@ See also technical docs related to this plugin:
 - [Data model](./model.md#prometheustimeseriesquery)
 - [Dashboard-as-Code Go lib](./go-sdk/query.md)
 
+## Annotation (`PrometheusPromQLAnnotation`)
+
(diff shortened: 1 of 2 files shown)


AntoineThebaud (approved):
Thanks! 

Please open a dedicated PR for the CUE model issue you found (nice catch).

PR #812, opened 2026-09-11
[ENHANCEMENT] Support responsive Stat Chart tile orientations

TridipDam:
# Description
This PR updates the Stat Chart plugin to support responsive stat tile orientations.
The layout dynamically adapts to the available panel width, allowing stat values to wrap into multiple rows instead of requiring a single horizontal row when many series are displayed.

This addresses the responsive layout requirement described in the main Perses issue:
https://github.com/perses/perses/issues/4382

# Screenshots
Screenshot 2026-09-11 141349

Screenshot 2026-09-11 145914

image-1

# Checklist

- [x] Pull request has a descriptive title and context useful to a reviewer.
- [x] Pull request title follows the `[<catalog_entry>] <commit message>` naming convention using one of the
  following `catalog_entry` values: `FEATURE`, `ENHANCEMENT`, `BUGFIX`, `BREAKINGCHANGE`, `DOC`,`IGNORE`.
- [x] All commits have [DCO signoffs](https://github.com/probot/dco#how-it-works).

## UI Changes

- [x] Changes that impact the UI include screenshots and/or screencasts of the relevant changes.
- [x] Code follows the [UI guidelines](https://github.com/perses/perses/blob/main/ui/ui-guidelines.md).

PR #814, opened 2026-09-12
[FEATURE] StatChart: configurable multi-series layout (auto/row/grid)

colivi:
Relates to https://github.com/perses/perses/issues/4382

# Description

Multi-value StatCharts laid all series on one horizontal row (overflow scroll).

Maximizes space wrt number of elements to fill so panels that have multiple sub-panels to display stay readable.

- auto: pick columns from series count (4→2×2, 6→3×2, …)

- row: legacy horizontal strip

- grid: wrap; optional fixed seriesColumns (2, 3, …)

- Options editor: Series layout control

# Screenshots
N/A until deployed on Argos; change is layout-only in "StatChartPanel.tsx".

# Checklist
- [x] "[FEATURE]" title
- [x] DCO signoff


shahrokni (commented):
⚠️ 
Please hold this one.
We have already a PR from another contributor being reviewed and is in final stage.

colivi:
Hi @shahrokni, 

OK we’ll hold #814 in favor of #812 for the moment.

When orientation is auto, does the layout wrap into multiple rows based on panel width and the number of series (keeping each tile at least a minimum readable width)?

=> That’s the behavior we need for dense multi-series stats.


We’ll re-test Argos multi-series stats after merge and if gaps remain we will open a follow-up.
Thanks

shahrokni:
> Hi @shahrokni,
> 
> OK we’ll hold #814 in favor of #812 for the moment.
> 
> When orientation is auto, does the layout wrap into multiple rows based on panel width and the number of series (keeping each tile at least a minimum readable width)?
> 
> => That’s the behavior we need for dense multi-series stats.
> 
> We’ll re-test Argos multi-series stats after merge and if gaps remain we will open a follow-up. Thanks

You can take a look at this output.
The very first one (in the screenshot) shows your requirement I think. If not, let me know
> When orientation is auto, does the layout wrap into multiple rows based on panel width and the number of series (keeping each tile at least a minimum readable width)?

statchart/src/stat-chart-model.ts
```diff
@@ -60,6 +73,36 @@ export interface StatChartOptions {
...
+  }
+  if (seriesLayout === 'grid' && seriesColumns !== undefined && seriesColumns !== null && seriesColumns >= 1) {
+    return Math.min(12, Math.floor(seriesColumns));
+  }
+  // auto (and grid without columns)
+  if (seriesCount <= 2) return seriesCount;
+  if (seriesCount <= 4) return 2;
+  if (seriesCount <= 6) return 3;
+  if (seriesCount <= 9) return 3;
+  return Math.ceil(Math.sqrt(seriesCount));
```
s3onghyun, line 105: Small thing: these `auto` branches all evaluate to `Math.ceil(Math.sqrt(seriesCount))` — I checked n=1..30 and they match exactly (2→2, 4→2, 6→3, 9→3, 10→4). So the ladder collapses to a single `return Math.ceil(Math.sqrt(seriesCount));`. If you kept the steps for readability that's fine; just flagging that none of them currently override the formula, so it's easy to drift if one is edited later. Not blocking.

PR #815, opened 2026-09-13
[FEATURE] Markdown: Preserve safe link targets and open external links in new tabs

colivi:
# Description

Markdown panels render links via "marked" + "DOMPurify". Absolute external URLs were opened in the same browser tab and replaced the Perses dashboard.

This change:

- Registers a one-time DOMPurify "afterSanitizeAttributes" hook so anchors get a safe new-tab target when appropriate
- Passes "ADD_ATTR: ['target', 'rel']" so those attributes survive sanitization
- Behavior:
  - absolute "http(s)://" (and protocol-relative "//") → "target="_blank"" + "rel="noopener noreferrer""
  - explicit "target="_blank"" (including relative URLs) → kept + "rel" set
  - other relative links (e.g. "/explore") → **same tab** (no forced blank)

Unit tests cover the three cases above ("MarkdownPanel.sanitize.test.ts").

# Screenshots

N/A for this PR (behavior-only). After deploy: Markdown “Quick links” with absolute URLs open in a new tab; in-app relative links stay in-place unless "target="_blank"" is set in the source.

# Checklist

- [x] Pull request title follows the "[<catalog_entry>] <commit message>" naming convention using one of the
  following "catalog_entry" values: "FEATURE", "ENHANCEMENT", "BUGFIX", "BREAKINGCHANGE", "DOC","IGNORE".

## UI Changes

- [x] Changes that impact the UI include screenshots and/or screencasts of the relevant changes. *(behavior-only; optional follow-up screenshot after plugin publish)*


prometheus/src/plugins/prometheus-time-series-query/plugin.test.ts
```diff
@@ -136,24 +136,6 @@ describe('PrometheusTimeSeriesQuery', () => {
     expect(results.series[0]?.formattedName).toEqual('bar - format');
   });
 
-  it('should evaluate instant queries at the exact end of the time range, not the step-aligned end', async () => {
```
jgbernalp, line 139: not sure why this was removed as this change should target only the markdown plugin. It seems there are other unrelated changes can you check?

colivi: Good catch, the branch was based on an older local commit that also carried unrelated Prometheus/table work. I’ve rebased onto main so this PR only contains the markdown plugin changes (external links target=_blank after DOMPurify). The Prometheus test is unchanged relative to main again.

markdown/src/MarkdownPanel.tsx
```diff
@@ -75,9 +75,42 @@ function markdownToHTML(text: string): string {
...
+ *
+ * DOMPurify may strip target without ADD_ATTR; the afterSanitizeAttributes hook
+ * re-applies attributes after sanitization.
+ */
+export function installMarkdownExternalLinkHook(): void {
+  if (externalLinkHookInstalled) {
+    return;
+  }
+  externalLinkHookInstalled = true;
+  DOMPurify.addHook('afterSanitizeAttributes', (node) => {
```
shahrokni: I guess, as soon as the `afterSanitizeAttributes` is registered, regardless of what exact plugin we are dealing with, the new behavior is applied everywhere, whenever an external link is clicked. Although the new introduced behavior makes sense for 100% of cases (imo), the main question is why we should do it specifically from markdown plugin? This later could cause confusion.
@jgbernalp  WDYT?

jgbernalp: @shahrokni good catch, this will mean all instances of dompurify would use the same hook. Since the DOMPurify export is a factory, a better approach would be to create an instance so the hook is isolated

```typescript
const markdownPurifier = DOMPurify(window);

markdownPurifier.addHook('afterSanitizeAttributes', (node) => { ... });

export function sanitizeHTML(html: string): string {
  return markdownPurifier.sanitize(html, {
...

```

shahrokni: @colivi

colivi: Thanks, agreed. Switched to an isolated DOMPurify(window) instance so the hook only applies to the markdown plugin.

shahrokni (approved):
Tested and approved!

Image

PR #820, opened 2026-09-17
[ENHANCEMENT] Pyroscope: scope profile type options to the selected service

luweglarz:
# Description

The profile type dropdown currently lists every profile type that exists in the tenant, regardless of the selected service. Picking one the service does not emit (for example a `memory:*` type on a CPU-only service) produces an empty flamegraph with no indication why.

Pyroscope's `ProfileTypes` RPC only accepts `start`/`end` and has no `matchers` field, so it cannot be filtered by service on its own. `LabelValues` does support matchers, and the profile type name is exposed as the `__name__` label, so the dropdown is now built from both calls:

1. `ProfileTypes` for the full catalog, which is the only source of the complete 
`<name>:<sample-type>:<sample-unit>:<period-type>:<period-unit>` IDs required by `profileTypeID` in `SelectMergeStacktraces` / `SelectSeries`.
2. `LabelValues` with `name: __name__` and `matchers: ['{service_name="<service>"}']` to get the profile type names that actually exist for that service.

# Checklist


pyroscope/src/utils/use-query.ts
```diff
@@ -69,20 +69,41 @@ export function useLabelValues(
...
-    queryKey: ['searchProfileTypes', client],
+    queryKey: ['searchProfileTypes', service, client],
     queryFn: async () => {
-      return await client!.searchProfileTypes(
+      const profileTypesResponse = await client!.searchProfileTypes(
         {},
         { 'content-type': 'application/json' },
         { start: start * MILLISECONDS, end: end * MILLISECONDS },
       );
+
```
AntoineThebaud, line 87: maybe a small comment here to explain the purpose of this second call as you did in the PR desc:
```suggestion
// when applicable, filter the list of profile types to keep only the ones relevant for the given service
```

AntoineThebaud (approved):
Appr with suggestion

Nexucis (approved):

PR #823, opened 2026-09-18
[ENHANCEMENT] PieChart: radius configuration added

dchromik:
# Description

Relates to https://github.com/perses/perses/issues/4426
- Changed radius to inner and outer radius, inner is optional
- Radius was hard coded at 50 but ignored in FE where 90% was used, new default value is outer radius 90% and no inner radius
- Moved radii and color palette to Visual group
- Added migration from flat radius and color palette
- Looks to me like piechart was copied from time series chart, there are some unused values, marked them as deprecated as deleting them would mean a breaking change

# Screenshots

image

# Checklist

## UI Changes


piechart/src/PieChartOptionsEditorSettings.tsx
```diff
@@ -39,79 +39,173 @@ import { CalculationSelector, LegendOptionsEditor } from '@perses-dev/plugin-sys
...
+}
+
+interface RadiusControlsProps {
+  innerRadius?: string;
+  outerRadius: string;
+  onChange: (radius: RadiusInputs) => void;
+}
+
+function RadiusControls({ innerRadius = '', outerRadius, onChange }: RadiusControlsProps): ReactElement {
+  const [isOuterRadiusEmpty, setIsOuterRadiusEmpty] = useState(false);
```
jgbernalp: it seems this state is not updated correctly when the prop changes, do we need this state or can we just check the prop?

dchromik: The state was not needed, good point. I folded the RadiusControls component into it's parent since it didn't really have a purpose after the state was removed.

piechart/schemas/pie.cue
```diff
@@ -30,7 +30,15 @@ spec: close({
 	format?:     common.#format
 	sort?:       "asc" | "desc"
 	mode?:       "value" | "percentage"
+
 	showLabels?: bool
-	radius:      number
+
+	// Deprecated top-level visual options remain accepted for persisted dashboards.
+	radius?: number
 	colorPalette?: [...string]
```
jgbernalp, line 38: is this colorPalette also deprecated and included inside visual?

piechart/src/PieChartOptionsEditorSettings.tsx
```diff
@@ -39,79 +39,116 @@ import { CalculationSelector, LegendOptionsEditor } from '@perses-dev/plugin-sys
...
+    onChange(
+      produce(value, (draft) => {
+        draft.visual.innerRadius = innerRadius || undefined;
+        draft.visual.outerRadius = outerRadius;
+      }),
+    );
+  }
+
+  function handleOuterRadiusChange(event: ChangeEvent<HTMLInputElement>): void {
+    updateRadius(value.visual.innerRadius ?? '', event.target.value);
```
jgbernalp: it seems we are not validating that the outer radius needs to b bigger than the inner one, still works because echarts handles the case, but it might produce unintended behavior

jgbernalp: Image

dchromik: I wasn't sure how strongly should the input be validated, the user can see immediately how it looks like.
It would be tricky to validate given that radii can be a mix of absolute px and relative % value. With panel resizing done outside of piechart editor we won't be able to tell which is bigger.
This could be addressed by:
1. Changing radii to a number and always passing relative values to ECharts
2. Adding unit toggle for radii, forcing both to be absolute or relative

piechart/schemas/pie.cue
dchromik: Yes, I'll update the comment to make it clearer.

piechart/src/PieChartOptionsEditorSettings.tsx
dchromik: The screenshot didn't load up at first for me, using only relative values and capping at 100% would be safest here and wouldn't require any heavy changes.

piechart/src/pie-chart-model.ts
```diff
@@ -42,8 +60,16 @@ export function createInitialPieChartOptions(): PieChartOptions {
...
     showLabels: false,
     sort: DEFAULT_SORT,
+    visual: { ...DEFAULT_VISUAL },
+  };
+}
+
+export function resolvePieChartVisualOptions(options: PieChartOptions): PieChartVisualOptions {
+  return {
+    innerRadius: options.visual?.innerRadius,
+    outerRadius: options.visual?.outerRadius ?? DEFAULT_OUTER_RADIUS,
```
jgbernalp, line 72: This should also catch empty strings

piechart/schemas/pie.cue
```diff
@@ -30,7 +30,15 @@ spec: close({
...
 	mode?:       "value" | "percentage"
+
 	showLabels?: bool
-	radius:      number
+
+	// Deprecated: radius and colorPalette moved to visual options, but remain accepted at the top level for persisted dashboards.
+	radius?: number
 	colorPalette?: [...string]
+	visual?: close({
+		innerRadius?: string
```
jgbernalp: can we enforce a more strict validation for example a regex including percentage or unitless numbers?

piechart/src/pie-chart-model.ts
dchromik: Radii changed to numbers, the check should be OK now

piechart/schemas/pie.cue
dchromik: Radii changed to numbers, also added check for inner < outer - 5. If inner was close to outer only a thin line rendered but we could only leave a simple inner < outer.

dchromik:
Sorry for sitting on this for so long, the rework was a bit larger than expected.
I think that the range slider is better than 2 numeric inputs for 0-90% values, screenshot in description updated.
If existing piechart from previous commits is broken recreate it, I didn't add normalization from strings.

piechart/schemas/migrate/migrate.cue
```diff
@@ -52,11 +52,13 @@ spec: {
...
-	}
+	visual: {
+		#colorMode: *#panel.fieldConfig.defaults.color.mode | null
+		if #colorMode == "shades" {
+			#mappedColor: *commonMigrate.#mapping.color[#panel.fieldConfig.defaults.color.fixedColor] | "#555555"
+			colorPalette: [#mappedColor]
+		}
 
-	radius: 50
+		outerRadius: 90
```
jgbernalp: Can this value be 100?
```suggestion
		outerRadius: 100
```

piechart/schemas/pie.cue
```diff
@@ -30,7 +30,18 @@ spec: close({
...
+
 	showLabels?: bool
-	radius:      number
+
+	// Deprecated: radius and colorPalette moved to visual options, but remain accepted at the top level for persisted dashboards.
+	radius?: number
 	colorPalette?: [...string]
+	visual?: close({
+		innerRadius?: int & >=0 & <=85
+		outerRadius:  int & >=5 & <=90
```
jgbernalp: Sorry I din't mean this to be too restrictive just wondering about the type, some users might want options like 99 and 100 so I think this should be open
```suggestion
		innerRadius?: int & >=0 & <=100
		outerRadius:  int & >=0 & <=100
```

piechart/schemas/pie.cue
```diff
@@ -30,7 +30,18 @@ spec: close({
...
+		if innerRadius != _|_ {
+			outerRadius: >=(innerRadius + 5)
+		}
```
jgbernalp:
```suggestion
```

piechart/schemas/migrate/migrate.cue
dchromik: Yes, when it is 100 the piechart gets clipped a bit, so only a small visual issue.

jgbernalp: is it then an issue with the default styles while rendering, I think users might be confused as why is not 100 (100%)

dchromik: Yes, I agree that it was confusing, the range is 0-100 now.
Here's the clipping, it's really minor: 
image

piechart/src/PieChartBase.tsx
```diff
@@ -31,6 +34,8 @@ registerECharts([
   LegendComponent,
   CanvasRenderer,
 ]);
+
+const CHART_SX = { width: '100%', height: '100%' };
```
jgbernalp, line 38: should we set this to 90% so there is no clipping and let the default value as 100?

dchromik: Updated styles, added padding to EChart config for piecharts.

piechart/sdk/go/pie.go
```diff
@@ -18,7 +18,10 @@ import (
 	"github.com/perses/perses/go-sdk/panel"
 )
 
-const PluginKind = "PieChart"
+const (
+	PluginKind         = "PieChart"
+	defaultOuterRadius = 90
```
jgbernalp: this can now be:

```suggestion
	defaultOuterRadius = 100
```

piechart/schemas/migrate/migrate.cue
jgbernalp: this can now be:

```suggestion
		outerRadius: 100
```

piechart/src/pie-chart-model.ts
```diff
@@ -19,6 +19,21 @@ import type { Definition } from '@perses-dev/spec';
 export const DEFAULT_FORMAT: FormatOptions = { unit: 'decimal', shortValues: true };
 export const DEFAULT_SORT: SortOption = 'desc';
 export const DEFAULT_MODE: ModeOption = 'value';
+export const DEFAULT_OUTER_RADIUS = 90;
```
jgbernalp:
```suggestion
export const DEFAULT_OUTER_RADIUS = 100;
```

jgbernalp (approved):

jgbernalp:
some unit tests are failing, mind checking @dchromik ?

PR #824, opened 2026-09-18
[IGNORE] Use spec in table schema

Nexucis:
I missed this plugin when moving from perses/perses to spec

relates perses/perses#4477


table/schemas/table.cue
```diff
@@ -16,7 +16,8 @@ package model
 import (
 	"strings"
 	"github.com/perses/shared/cue/common"
-	model "github.com/perses/perses/cue/model/api/v1/common"
+	model "github.com/perses/spec/cue/common"
```
AntoineThebaud: looks like this line can be removed?

Nexucis: good catch !

celian-garcia (approved):
<3

